PR #19, opened 2025-06-28
Fix timezone handling bug in timestamp parsing

nishantmodak:
- Parse ISO timestamps as UTC instead of local server time
- Ensures consistent behavior across different server timezones
- Fixes bug where queries returned logs from wrong time periods
- Maintains lookback_minutes functionality when only start_time provided
- Add comprehensive test suite covering all timestamp scenarios

Resolves issue where "2025-06-23 16:00:00" query returned June 24th logs due to timezone offset confusion.


akshaynanavare (approved):
LGTM 🚀

PR #122, opened 2026-04-06
Fix exception investigation: get_service_traces instead of get_traces

prathamesh-sonpatki:
## Summary

- `get_exceptions` description was misleading: claimed to return `trace_id` and stack traces, but both are always null/empty (data comes from aggregated metrics)
- Claude was defaulting to `get_traces` (complex JSON pipeline) when investigating exceptions, instead of the simpler `get_service_traces(service_name)` — causing repeated tool calls and timeouts
- No explicit stop condition: Claude kept calling additional tools (`get_service_logs`, `get_traces`) after already finding traces

## Changes

**`internal/telemetry/traces/description.go`**
- Corrects description: `trace_id` is always null, no stack traces available
- Adds explicit 3-step investigation flow with a hard stop after `get_service_traces`
- Adds `deployment_environment` → `env` propagation rule for follow-up calls
- Adds rule: if trace_id is available, use `get_service_traces(trace_id)` not `get_traces`

**`internal/prompts/descriptions/get_traces.md`**
- Adds rule: when navigating from `get_exceptions` results, use `get_service_traces` — not `get_traces` with a JSON pipeline

**`internal/prompts/descriptions/exception_investigation.md`** *(new)*
- Combined description used by the eval harness (`~/Projects/last9-mcp-evals`) to test exception investigation behaviour

**`CLAUDE.md`** *(new)*
- Documents eval harness location and the description file ↔ eval operation mapping

## Eval results

| Case | What it tests | Before | After |
|------|--------------|--------|-------|
| eq-003 | Full investigation flow must not use `get_traces` | ✗ fail | ✓ pass |
| eq-004 | Null trace_id — must use `get_service_traces` | ✗ fail | ✓ pass |
| eq-005 | Stop after finding traces — no additional tool calls | ✗ fail | ✓ pass |
| eq-006 | `deployment_environment` propagated as `env` | ✗ fail | ✓ pass |

```
Before: 3/7 passed (4 failed)
After:  7/7 passed (0 failed)
```

## Summary by CodeRabbit

* **Documentation & Improvements**
  * Updated `get_exceptions` tool instructions with clarified workflow for retrieving aggregated server exceptions and related traces
  * Enhanced guidance for transitioning from exception results to trace lookups


rizwan2000rm:
@coderabbitai full review

rizwan2000rm (approved):

PR #148, opened 2026-05-22
feat: resolve indicator PromQL inline in alert config tools

prathamesh-sonpatki:
## Summary

- `get_alert_config` now resolves each rule's `expression_args` indicators to their actual PromQL queries via `GET /entities/{id}/kpis/{id}`, with result caching by `(entityID, kpiID)` to avoid redundant fetches
- New `get_entity_alert_rules` tool fetches entity-scoped alert rules (which includes `expression_args` in the SQL query, unlike the org-wide endpoint) and resolves PromQL inline
- `loadTestEnv()` walks up the directory tree to find `.env` at the project root
- Formatted output is deterministic: indicator names, variable keys, and property keys are sorted before rendering

## Background

The org-wide `GET /alert-rules` endpoint omits `expression_args` from its SQL SELECT. The entity-scoped `GET /entities/{id}/alert-rules` includes it. The new tool uses the entity-scoped endpoint; the existing tool's KPI resolution now triggers when rules do have `expression_args` populated (e.g. rules fetched after switching to entity-scoped calls).

## Test plan

- [ ] Unit tests: `go test ./internal/alerting/...` — all pass
- [ ] Integration tests (requires `.env`): `go test -run Integration ./internal/alerting/` — verified live against org, 7 rules returned with resolved PromQL for entity `0466b646`
- [ ] No secrets in diff — verified with `git diff` grep scan
- [ ] Deterministic output verified: sorted map iteration in formatter

### DIFF
--- a/internal/alerting/alerts.go
+++ b/internal/alerting/alerts.go
@@ -104,8 +104,9 @@ type AlertInstance struct {
 
 const GetAlertConfigDescription = `
 	Get alert configurations (alert rules) from Last9.
-	Returns configured alert rules and supports both typed filters and free-text search.
-	Uses the datasource configured in the server config (or default if not specified).
+	Returns configured alert rules with metadata and supports both typed filters and free-text search.
+	Use this tool first to discover rules and entity IDs, then call get_entity_alert_rules with an
+	entity_id to get the full expression logic and resolved PromQL for that entity's rules.
 
 	Optional filters:
 	- rule_id: Exact match on alert rule ID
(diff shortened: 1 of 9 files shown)


internal/alerting/alerts.go
```diff
@@ -89,8 +104,9 @@ type AlertInstance struct {
 
 const GetAlertConfigDescription = `
 	Get alert configurations (alert rules) from Last9.
-	Returns configured alert rules and supports both typed filters and free-text search.
-	Uses the datasource configured in the server config (or default if not specified).
+	Returns configured alert rules with metadata and supports both typed filters and free-text search.
+	Use this tool first to discover rules and entity IDs, then call get_entity_alert_rules with an
+	entity_id to get the full expression logic and resolved PromQL for that entity's rules.
```
adityagodbole: "then if required, use get_entity_details with an
	entity_id to get the PromQL for the indicator and other details of the alert group (entity) of the alert rule"

PR #159, opened 2026-06-01
feat: alert rule state filters and grouping

adityagodbole:
# Alert Rule State Tool Updates

## Changes Made
- Upgraded the `get_alert_rule_state` tool's schema to accept multiple filters: `alert_group_id`, `rule_name`, `alert_group_name`, `label_filters`, and `state`.
- Ensured strict 1:1 mapping with the upstream Last9 `/alerts/monitor` API, providing all supported native filters while omitting synthetic client-side filters (such as `alert_rule_id`).
- Remapped the historical timeseries JSON output so it groups datapoints explicitly by `rule_id` (returning `map[string][]Datapoint`), providing structured context instead of a flat array.
- Updated the MCP JSON-RPC schemas and verified `invopop/jsonschema` reflection tag formatting.
- Fixed a panic on startup during tool registration by removing `WORD=` formatted prefixes like `required,description=` from struct tags.

## Validation Results
- Verified correct map-based JSON extraction and test cases natively handling all returned rules in `internal/alerting/alert_rule_state_test.go`.
- Successfully built `last9-mcp` binary (`go build .`).
- Executed the actual MCP server native binary (`go run .`) with a wrapper client script sending JSON-RPC payloads over `stdin`.
- Captured a fully functioning response indicating a `2.0` JSON-RPC success payload from the tool containing the correctly grouped timeseries maps across multiple `rule_id`s.

## Design Decisions
- **Filter Mapping Parity**: Refactored the tool to only provide filters natively supported by the Last9 API, strictly mapping the query parameters directly without applying any post-processing pruning.
- **Refresh Tokens**: Added notes to the project's knowledge base (`gemini-local.md`) highlighting that the Last9 API utilizes refresh tokens that logically never expire, but short-lived access tokens require frequent exchange.
- **Strict JSON Schemas**: Refactored `jsonschema` struct tags to use the standardized string-only description format since the `mcp-go-sdk` wrapper rigidly panics on assignment operators (`=`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `get_alert_rule_state` tool for historical firing state (1/0) per alert rule over a time range, grouped by `rule_id`. Supports server-side filtering by `alert_group_id`, `rule_name`, `alert_group_name`, `label_filters`, and `state` (#159).
+
 ## [0.7.5] - 2026-06-01
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


prathamesh-sonpatki (approved):
All review feedback addressed: cap math corrected, duplicate auth header removed, validation paths return IsError, filter forwarding asserted, e2e integration test verifies live backend behavior. CI green. Approving and merging.

PR #179, opened 2026-07-06
feat(prompts): conditional STOP in get_exceptions + aggregate-first investigation guidance in get_logs

prathamesh-sonpatki:
## Problem

`get_exceptions.md` carried an unconditional "STOP — do not call log tools" instruction. This misdirected investigations on log-heavy/severity-less services, where span exceptions are downstream symptoms (retry storms, pool timeouts) and the root cause exists only in log bodies.

## Changes

**get_exceptions.md**: rewritten as a regime conditional instead of a blanket stop. Trace-instrumented services still stop (timeout protection for broad raw pulls retained). Log-heavy services continue to logs via aggregate/count pipelines (cheap, wide-window-safe). Regime discriminator: `physical_index_service_count` log presence.

**get_logs.md**: new CRITICAL block for incident investigation:
- Discover error signatures via ERROR-gated group-by (never guess strings)
- Never count a bare component/logger name without the ERROR gate (observed ~750x overcount in the wild)
- Trend/onset via `window_aggregate`
- Sanity-check counts against the service's APM error rate
- Fetch raw exemplars last

**.gitignore**: adds `.codex/` and other local agent/tool dirs so local credentials and agent state can never be committed.

## Eval evidence (last9-mcp-evals)

- Exception suite: 9/9 passing with the new text vs 7/9 with the old (eq-008/eq-009 fail on the old misdirecting STOP)
- Log suite: passes including new lq-104/lq-105 count-discipline cases

Pairs with the last9-mcp-evals PR `feat/investigation-guardrail-evals`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #180, opened 2026-07-07
feat(logs): deterministic guardrails — count sanity ratio + self-healing pipeline errors

prathamesh-sonpatki:
## Summary

Two deterministic guardrails on the `get_logs` JSON-pipeline path, complementing the prose guidance already shipped in the tool descriptions (#179) and the last9-mcp-evals#27 evals — per the enforcement-ladder decision, guidance that can be tool behavior should be tool behavior.

### U3 — count-vs-volume sanity ratio (`internal/utils/count_sanity.go`)
When a `get_logs` pipeline is an ERROR/count-style aggregation, the response gets a machine-readable `l9_sanity` block so an over-broad filter is self-evident, instead of silently returning a wildly inflated count.

Fires only when **all** hold, otherwise silently skipped:
- The pipeline has an `aggregate`/`window_aggregate` stage whose function(s) include `$count`.
- Exactly one service is extractable from the filter stages via an `$eq` on `ServiceName` (searched recursively through `$and`/`$or`/`$not`). Zero or multiple distinct services → skip.
- The aggregate response's matched count is parseable (sum of the group/bucket count values).

Baseline: one Prometheus instant query against `sum(sum_over_time(physical_index_service_count{service_name="<svc>"}[<window>]))`, evaluated at the window end, using the same helper (`MakePromInstantAPIQuery`) and datasource credentials as other prom-backed tools.

Output — attached at the top level of the tool response:
```json
"l9_sanity": {
  "matched_count": 750,
  "service_log_volume": 1000,
  "ratio": 0.75,
  "note": "matched count is 75.00% of ALL log lines this service emitted in the window — if this was meant to count errors, the filter is likely too broad (e.g. matching a component/logger name without an ERROR gate); re-narrow and re-count."
}
```
`note` is `""` when `ratio <= 0.05` (numbers are still included). No hard threshold, no blocking — informational only.

Failure semantics: baseline query error/non-200/unparseable response, or a count that can't be summed → the guardrail returns the original response completely unchanged. No retries, no added latency beyond the one instant query.

### U5 — self-healing pipeline-validation errors (`internal/utils/service_logs.go`)
When the backend log API rejects a pipeline with HTTP 400, `MakeLogsJSONQueryAPI` now wraps the error with the pipeline stage schema (`type` enum, filter example, exact field names, parse-before-Body-derived-fields, and a pointer to `get_log_attributes_for_pipeline`) appended after the original status + body. The client learns the fix in-band instead of getting a bare status. Only 400s get the hint — other statuses are returned unchanged for the caller to handle as before.

## Test coverage
- `internal/utils/count_sanity_test.go`: high ratio → note present; ratio ≤5% → numbers present, empty note; no count aggregate → untouched (no HTTP call); two distinct `ServiceName` values → skipped (no HTTP call); baseline prom query error → untouched.
- `internal/utils/service_logs_test.go`: 400 → error contains hint + original body; 500 → response returned unchanged, no error; 200 → no error.
- `internal/telemetry/logs/query_sanitize_test.go`: updated the existing `TestGetLogsHandlerNormalizesAliasesBeforeAPICall` fixture server to tolerate the guardrail's second (baseline) request at a different endpoint path, and to expect 2 requests instead of 1 — this test's pipeline happens to match the U3 fire condition (single `ServiceName` + `$count` aggregate).

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/utils/... ./internal/telemetry/logs/...`
- [x] `go test ./...` (full repo, unaffected packages still green)

### DIFF
--- a/internal/utils/count_sanity.go
+++ b/internal/utils/count_sanity.go
@@ -1,0 +1,243 @@
+package utils
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"math"
+	"net/http"
+
+	"last9-mcp/internal/models"
+)
+
+// HasCountAggregateStage reports whether pipeline contains an aggregate or
+// window_aggregate stage whose function/aggregates include a $count.
+func HasCountAggregateStage(pipeline []map[string]interface{}) bool {
+	for _, stage := range pipeline {
+		stageType, _ := stage["type"].(string)
+
+		switch stageType {
+		case "aggregate":
+			aggregates, ok := stage["aggregates"].([]interface{})
+			if !ok {
+				continue
+			}
+			for _, rawAggregate := range aggregates {
+				aggregate, ok := rawAggregate.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				if isCountFunction(aggregate["function"]) {
+					return true
+				}
+			}
+		case "window_aggregate":
+			if isCountFunction(stage["function"]) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func isCountFunction(rawFunction interface{}) bool {
+	function, ok := rawFunction.(map[string]interface{})
+	if !ok {
+		return false
+	}
+	_, hasCount := function["$count"]
+	return hasCount
+}
+
+// ExtractSingleServiceName scans a pipeline's filter stages (including inside
+// $and/$or/$not) for $eq conditions on the ServiceName field. It returns the
+// value and true only when exactly one distinct service name is present.
+func ExtractSingleServiceName(pipeline []map[string]interface{}) (string, bool) {
+	services := map[string]struct{}{}
+
+	for _, stage := range pipeline {
+		stageType, _ := stage["type"].(string)
+		if stageType != "filter" {
+			continue
+		}
+		collectServiceNameEquals(stage["query"], services)
+	}
+
+	if len(services) != 1 {
+		return "", false
+	}
+	for service := range services {
+		return service, true
+	}
+	return "", false
+}
+
+func collectServiceNameEquals(condition interface{}, out map[string]struct{}) {
+	switch typed := condition.(type) {
+	case map[string]interface{}:
+		for key, value := range typed {
+			switch key {
+			case "$eq":
+				args, ok := value.([]interface{})
+				if !ok || len(args) != 2 {
+					continue
+				}
+				field, ok := args[0].(string)
+				if !ok || field != "ServiceName" {
+					continue
+				}
+				if service, ok := args[1].(string); ok {
+					out[service] = struct{}{}
+				}
+			case "$and", "$or", "$not":
+				collectServiceNameEquals(value, out)
+			}
+		}
+	case []interface{}:
+		for _, item := range typed {
+			collectServiceNameEquals(item, out)
+		}
+	}
+}
+
+// SumAggregateCount sums the count values across the groups/buckets of a log
+// aggregate API response. It returns false when the response can't be
+// parsed as an aggregate result (guardrail must skip, not fail, in that case).
+func SumAggregateCount(response map[string]interface{}) (float64, bool) {
+	data, ok := response["data"].(map[string]interface{})
+	if !ok {
+		return 0, false
+	}
+	result, ok := data["result"].([]interface{})
+	if !ok {
+		return 0, false
+	}
+	if len(result) == 0 {
+		return 0, true
+	}
+
+	var sum float64
+	found := false
+	for _, rawItem := range result {
+		item, ok := rawItem.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		if point, ok := item["value"].([]interface{}); ok && len(point) == 2 {
+			sum += promNumberFromAny(point[1])
+			found = true
+		}
+		if points, ok := item["values"].([]interface{}); ok {
+			for _, rawPoint := range points {
+				point, ok := rawPoint.([]interface{})
+				if !ok || len(point) != 2 {
+					continue
+				}
+				sum += promNumberFromAny(point[1])
+				found = true
+			}
+		}
+	}
+
+	if !found {
+		return 0, false
+	}
+	return sum, true
+}
+
+func promNumberFromAny(raw interface{}) float64 {
+	switch value := raw.(type) {
+	case float64:
+		return value
+	case json.Number:
+		f, err := value.Float64()
+		if err == nil {
+			return f
+		}
+	case string:
+		var f float64
+		if _, err := fmt.Sscanf(value, "%g", &f); err == nil {
+			return f
+		}
+	}
+	return 0
+}
+
+type promInstantVectorPoint struct {
+	Value []interface{} `json:"value"`
+}
+
+// AppendCountSanity attaches a top-level "l9_sanity" block to response
+// comparing a $count aggregate's matched_count against the service's total
+// log volume over the same window (physical_index_service_count). It is a
+// pure guardrail add-on: on any failure to determine a single service, parse
+// the matched count, or fetch/parse the baseline, it returns response
+// unchanged. Never blocks or alters the underlying result.
+func AppendCountSanity(ctx context.Context, client *http.Client, cfg models.Config, pipeline []map[string]interface{}, startMs, endMs int64, response map[string]interface{}) map[string]interface{} {
+	if !HasCountAggregateStage(pipeline) {
+		return response
+	}
+
+	service, ok := ExtractSingleServiceName(pipeline)
+	if !ok {
+		return response
+	}
+
+	matchedCount, ok := SumAggregateCount(response)
+	if !ok {
+		return response
+	}
+
+	windowMinutes := (endMs - startMs) / 60000
+	if windowMinutes < 1 {
+		windowMinutes = 1
+	}
+	promql := fmt.Sprintf(`sum(sum_over_time(physical_index_service_count{service_name=%q}[%dm]))`, service, windowMinutes)
+
+	resp, err := MakePromInstantAPIQuery(ctx, client, promql, endMs/1000, cfg)
+	if err != nil {
+		return response
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return response
+	}
+
+	var instant []promInstantVectorPoint
+	if err := json.NewDecoder(resp.Body).Decode(&instant); err != nil {
+		return response
+	}
+
+	var volume float64
+	found := false
+	for _, point := range instant {
+		if len(point.Value) != 2 {
+			continue
+		}
+		volume += promNumberFromAny(point.Value[1])
+		found = true
+	}
+	if !found || volume <= 0 {
+		return response
+	}
+
+	ratio := math.Round((matchedCount/volume)*10000) / 10000
+
+	note := ""
+	if ratio > 0.05 {
+		note = fmt.Sprintf(
+			"matched count is %.2f%% of ALL log lines this service emitted in the window — if this was meant to count errors, the filter is likely too broad (e.g. matching a component/logger name without an ERROR gate); re-narrow and re-count.",
+			ratio*100,
+		)
+	}
+
+	response["l9_sanity"] = map[string]interface{}{
+		"matched_count":      int64(math.Round(matchedCount)),
+		"service_log_volume": volume,
+		"ratio":              ratio,
+		"note":               note,
+	}
+
+	return response
+}
--- a/internal/utils/count_sanity.go
+++ b/internal/utils/count_sanity.go
@@ -101,8 +101,13 @@ func collectServiceNameEquals(condition interface{}, out map[string]struct{}) {
 }
 
 // SumAggregateCount sums the count values across the groups/buckets of a log
-// aggregate API response. It returns false when the response can't be
-// parsed as an aggregate result (guardrail must skip, not fail, in that case).
+// aggregate API response. Rows are shaped {"metric": {...}, "values": []} —
+// "metric" holds the aggregate's `as`-aliased count(s) as JSON numbers
+// alongside any group-by labels/`__ts__` as strings; "values" is unused here
+// (that's a Prometheus instant-query shape, not this API's). Any numeric
+// field in "metric" counts; group-by/`__ts__` strings are skipped. It returns
+// false when no row yields a numeric field (guardrail must skip, not
+// miscount, in that case).
 func SumAggregateCount(response map[string]interface{}) (float64, bool) {
 	data, ok := response["data"].(map[string]interface{})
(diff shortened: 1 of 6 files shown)


prathamesh-sonpatki:
**MCP evals** — **153/174 passed** ✗ (381s)

| Suite | Passed | Failed | |
|---|---|---|---|
| attribute_discovery | 0/5 | 5 | ✗ |
| change_events | 4/4 | 0 | ✓ |
| dashboard | 1/5 | 4 | ✗ |
| database_investigation | 0/5 | 5 | ✗ |
| datasource_selection | 4/4 | 0 | ✓ |
| did_you_mean | 4/4 | 0 | ✓ |
| drop_rules | 2/4 | 2 | ✗ |
| entity_alert_rules | 3/3 | 0 | ✓ |
| exception_investigation | 7/7 | 0 | ✓ |
| get_alert_config | 6/6 | 0 | ✓ |
| get_alert_rule_state | 5/5 | 0 | ✓ |
| get_alerts | 6/6 | 0 | ✓ |
| get_metrics | 3/3 | 0 | ✓ |
| log_query | 40/40 | 0 | ✓ |
| log_tool_selection | 13/13 | 0 | ✓ |
| notification_channels | 3/3 | 0 | ✓ |
| service_investigation | 0/5 | 5 | ✗ |
| service_logs_args | 1/1 | 0 | ✓ |
| service_traces_args | 5/5 | 0 | ✓ |
| trace_query | 23/23 | 0 | ✓ |
| type_coercion | 23/23 | 0 | ✓ |

### Failures

| Case | Operation | Rules |
|---|---|---|
| `atd-001` | attribute_discovery | valid_step_tools, must_contain |
| `atd-002` | attribute_discovery | must_contain |
| `atd-003` | attribute_discovery | valid_step_tools, must_contain, must_contain |
| `atd-004` | attribute_discovery | valid_step_tools, must_contain |
| `atd-005` | attribute_discovery | must_contain |
| `dash-002` | dashboard | must_contain, must_contain |
| `dash-003` | dashboard | must_contain, must_contain |
| `dash-004` | dashboard | must_contain, must_contain |
| `dash-005` | dashboard | must_contain, must_contain |
| `dbi-001` | database_investigation | valid_step_tools, must_contain |
| `dbi-002` | database_investigation | valid_step_tools, must_contain, must_contain |
| `dbi-003` | database_investigation | valid_step_tools, must_contain |
| `dbi-004` | database_investigation | valid_step_tools, must_contain, must_contain |
| `dbi-005` | database_investigation | valid_step_tools, must_contain |
| `dru-002` | drop_rules | must_contain, must_contain |
| `dru-004` | drop_rules | must_contain, must_contain |
| `si-001` | service_investigation | valid_step_tools, must_contain |
| `si-002` | service_investigation | must_contain, must_contain |
| `si-003` | service_investigation | must_contain, must_contain |
| `si-004` | service_investigation | must_contain, must_contain |
| `si-005` | service_investigation | valid_step_tools, must_contain, must_contain |

Evaluated against SHA `9a4db2c74387b9021affd40b0d60dc8fbf9d1650`.

internal/utils/count_sanity.go
```diff
@@ -0,0 +1,245 @@
...
+	for _, rawItem := range result {
+		item, ok := rawItem.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		metric, ok := item["metric"].(map[string]interface{})
+		if !ok {
+			continue
+		}
+		for _, value := range metric {
```
faiz9068: **Important — mixed-aggregate count pollution.** This loop sums *every* numeric field in `metric`, but `HasCountAggregateStage` fires on *any* `$count` even when the same `aggregate` stage also computes `$avg`/`$sum`/`$max`. A row like `{"_count":750,"avg_dur":123.5}` sums to **873.5**, not 750 (reproduced) — so `matched_count`/`ratio`/`note` are wrong. It never blocks or alters results, but a misleading sanity note defeats the guardrail's purpose.

**Fix:** collect the `$count` aggregate's `as` alias(es) from the `aggregate` stage and sum only those keys in `metric`, instead of "every number counts." This also removes the brittle dependency on all non-count labels (incl. `__ts__`) always arriving as JSON strings.

internal/utils/count_sanity.go
```diff
@@ -0,0 +1,245 @@
...
+					continue
+				}
+				field, ok := args[0].(string)
+				if !ok || field != "ServiceName" {
+					continue
+				}
+				if service, ok := args[1].(string); ok {
+					out[service] = struct{}{}
+				}
+			case "$and", "$or", "$not":
```
faiz9068: **Important — `$not`-negated service treated as a positive pin.** Recursing into `$not` here means a filter meaning `ServiceName != orders-service` still extracts `orders-service` as *the* single service (reproduced). The baseline volume query then targets the excluded service and the ratio is nonsensical.

**Fix:** don't recurse into `$not` when collecting positive equalities — or, if a `ServiceName $eq` is found under a `$not`, skip the guardrail entirely.

internal/utils/count_sanity.go
```diff
@@ -0,0 +1,245 @@
...
+func SumAggregateCount(response map[string]interface{}) (float64, bool) {
+	data, ok := response["data"].(map[string]interface{})
+	if !ok {
+		return 0, false
+	}
+	result, ok := data["result"].([]interface{})
+	if !ok {
+		return 0, false
+	}
+	if len(result) == 0 {
```
faiz9068, line 162: **Minor** — an empty `result` returns `(0, true)`, so `AppendCountSanity` still fires a Prometheus baseline query and attaches `l9_sanity` with `matched_count:0, ratio:0`. Harmless but an avoidable request; consider short-circuiting when the count is 0.

internal/utils/count_sanity.go
```diff
@@ -0,0 +1,245 @@
...
+		return response
+	}
+
+	windowMinutes := (endMs - startMs) / 60000
+	if windowMinutes < 1 {
+		windowMinutes = 1
+	}
+	promql := fmt.Sprintf(`sum(sum_over_time(physical_index_service_count{service_name=%q}[%dm]))`, service, windowMinutes)
+
+	resp, err := MakePromInstantAPIQuery(ctx, client, promql, endMs/1000, cfg)
```
faiz9068: **Minor** — this baseline query uses the inbound `ctx` directly, unlike the chunked log path which wraps calls in `constants.PerChunkHTTPTimeout`. A slow metrics backend adds latency directly to `get_logs`. Consider bounding this one instant query with its own timeout.

internal/utils/count_sanity.go
```diff
@@ -0,0 +1,245 @@
...
+	if !ok {
+		return response
+	}
+
+	matchedCount, ok := SumAggregateCount(response)
+	if !ok {
+		return response
+	}
+
+	windowMinutes := (endMs - startMs) / 60000
```
faiz9068, line 245: **Minor** — sub-minute windows floor to 1 whole minute (e.g. 90s → `[1m]`), a slight mismatch vs the log query window. Acceptable for a ratio, just noting it.

internal/utils/count_sanity.go
```diff
@@ -0,0 +1,245 @@
...
+	case float64:
+		return value
+	case json.Number:
+		f, err := value.Float64()
+		if err == nil {
+			return f
+		}
+	case string:
+		var f float64
+		if _, err := fmt.Sscanf(value, "%g", &f); err == nil {
```
faiz9068: **Minor** — `fmt.Sscanf(value, "%g", &f)` works, but `strconv.ParseFloat` is idiomatic, faster, and matches existing usage (`apm.go:208`).

internal/utils/count_sanity_test.go
```diff
@@ -0,0 +1,241 @@
+package utils
```
faiz9068, line 1: **Test gaps (should fix).** Good happy-path + main skip coverage, but missing: `$or`/`$not` traversal and the negated-service case (finding #2); same-service-twice dedup (→ should fire); the mixed-aggregate pollution case (finding #1); the exact `ratio == 0.05` boundary (plan says note must be `""`); and the real string-shaped Prometheus value (`"value":["ts","1000"]`) — tests only send numbers, so `promNumberFromAny`'s string branch is never exercised end-to-end. A PromQL-injection test (service name containing `"`) would also lock in the `%q` escaping.

faiz9068 (commented):
Automated review (read-only checkout, `go build`/`vet`/tests green at HEAD; the two Important findings below were reproduced with throwaway probes).

**Verdict: merge with fixes.** The core safety contract holds — U3 is strictly informational, returns the original response untouched on every early-return path, and U5 is scoped to HTTP 400 with the original body preserved. No Critical issues. But two confirmed logic bugs make the `l9_sanity` block actively misleading in reachable pipelines, which is worse than emitting nothing — worth fixing before merge. Inline comments below.

internal/utils/count_sanity.go
prathamesh-sonpatki: > mixed-aggregate count pollution

Addressed: we now extract the  aliases from the pipeline and sum **only** those keys from each row's  (so other numeric aggregates/labels can't pollute ).

internal/utils/count_sanity.go
prathamesh-sonpatki: > -negated service treated as a positive pin

Addressed: service extraction no longer recurses into , so  won't incorrectly pin  for the baseline query.

internal/utils/count_sanity.go
prathamesh-sonpatki: > an empty  returns  ... still fires a Prometheus baseline query

Addressed: guardrail now skips when matched_count can't be parsed *or* is 0/empty, avoiding the extra PromQL call and  noise.

internal/utils/count_sanity.go
prathamesh-sonpatki: > baseline query uses the inbound  directly ... Consider bounding this one instant query with its own timeout.

Addressed: baseline instant query now runs under  so a slow metrics backend can't stall .

internal/utils/count_sanity.go
prathamesh-sonpatki: > sub-minute windows floor to 1 whole minute ...

Yep — we intentionally floor to 1m here since the baseline metric is minute-granularity-ish and this is only used for a coarse ratio sanity signal.

internal/utils/count_sanity.go
prathamesh-sonpatki: >  ...  is idiomatic

Addressed: switched to  for the string-number branch.

internal/utils/count_sanity_test.go
prathamesh-sonpatki: > Test gaps (should fix).

Addressed: added coverage for the  negated-service case, same-service dedup, mixed-aggregate pollution, ratio==0.05 boundary, Prom value string shape, and service-name escaping in PromQL.

prathamesh-sonpatki:
Addressed the review findings:\n\n- Sum only  aliases for matched_count (avoid mixed-aggregate pollution)\n- Don't treat ServiceName under  as a positive pin\n- Skip guardrail on empty/zero matched counts\n- Bound baseline PromQL call with PerChunkHTTPTimeout\n- Expanded tests for edge cases + escaping

faiz9068 (approved):
## Re-review — all findings addressed ✅

Re-reviewed the delta since my last review (`9a4db2c..2c13deba`, commit `fix(logs): correct count-sanity aliasing and edge cases`). Verified at head `2c13deba` in a throwaway worktree: `go build`/`go vet`/`go test ./internal/utils/... ./internal/telemetry/logs/...` all pass, and CI is green (Run Tests, CodeQL, all analyzers).

### Prior findings — status

| # | Finding | Status | Fix mechanism (verified) |
|---|---------|--------|--------------------------|
| Important 1 | Mixed-aggregate count pollution | ✅ Resolved | New `ExtractCountAliases` collects each `$count` aggregate's `as` alias; `SumAggregateCount` now sums **only** those keys. `TestAppendCountSanity_MixedAggregateMetricSumsOnlyCountAlias` proves it: metric `{_count:750, avg_dur:123.5, ignored:999, also_str:"1"}` → `matched_count == 750`, not the polluted sum. |
| Important 2 | `$not`-negated service treated as positive pin | ✅ Resolved | `$not` dropped from the recursion (now `$and`/`$or` only), so equalities under a negation are ignored. `TestExtractSingleServiceName_NotNegatedServiceSkips` asserts a lone `$not{$eq ServiceName …}` yields no single service → guardrail skips. |
| Minor | Wasted baseline call on 0 matches | ✅ Resolved | `len(result)==0` → `(0,false)` and an explicit `matchedCount <= 0` early return. `TestAppendCountSanity_ZeroMatchedSkipsAndNoPromCall` confirms **zero** Prometheus calls. |
| Minor | No timeout on baseline query | ✅ Resolved | Wrapped in `context.WithTimeout(ctx, constants.PerChunkHTTPTimeout)`. |
| Minor | `Sscanf` vs `ParseFloat` | ✅ Resolved | Now `strconv.ParseFloat`. |
| Minor | Sub-minute window floor | Acknowledged | Left as-is (acceptable for a ratio). |
| Test gaps | `$not`/`$or`, dedup, mixed-aggregate, `0.05` boundary, string-shaped prom value, injection | ✅ Resolved | All added and non-vacuous — incl. `RatioBoundaryExactly5PctHasEmptyNote` (ratio 0.05 → empty note), `PromValueStringShape` (`value:["ts","1000"]` exercises the string branch end-to-end), `PromQueryEscapesServiceName` (asserts `service_name="orders\"service"`). |

### Mechanism scrutiny (new-same-class)
The `$not` fix ignores equalities nested under a negation — correct, since anything under `$not` is a negation, not a positive pin. `ExtractCountAliases` reads the identical stage shape as `HasCountAggregateStage`, so the two stay consistent. One benign residual: a `$count` aggregate with no `as` alias makes the guardrail skip (fail-closed) — safe for an informational-only feature, not a blocker.

**Verdict: Approve.** All Important findings fixed with verified mechanisms and tests; core safety contract intact.

internal/utils/count_sanity.go
faiz9068: Verified fixed: `ExtractCountAliases` + `SumAggregateCount` now sums only the $count `as` keys. `TestAppendCountSanity_MixedAggregateMetricSumsOnlyCountAlias` proves `matched_count == 750` against a metric that also carries `avg_dur`/`ignored`/`also_str`. ✅

internal/utils/count_sanity.go
faiz9068: Verified fixed: `$not` removed from the recursion, so negated equalities are no longer treated as a positive pin. `TestExtractSingleServiceName_NotNegatedServiceSkips` confirms a lone `$not{$eq ServiceName …}` → no single service → guardrail skips. ✅

PR #181, opened 2026-07-08
feat(logs): detect logfmt + plaintext-inline Body fields in get_log_attributes_for_pipeline

prathamesh-sonpatki:
## Problem

`get_log_attributes_for_pipeline`'s Body-derived field discovery only `json.Unmarshal`s each
sampled log line. Lines that aren't JSON fail the unmarshal and are silently skipped — so for
plaintext/logfmt-logging services (the majority of a real fleet, ~86%), the tool returns ZERO
body-derived fields even though the logs clearly contain a level and other fields, and the
product UI (which does parse plaintext) shows them.

## Fix

In the Body-derived sampling path:
1. JSON path unchanged and tried first (existing behavior for JSON bodies).
2. On a non-JSON line, fall back to **logfmt** detection: if the line has 2+ distinct
   `key=value` pairs, extract those keys with a `{"type":"parse","parser":"logfmt","field":"Body"}`
   hint.
3. If not logfmt either, fall back to **plaintext-inline** severity detection: a
   timestamp-anchored pattern (`\d{2}:\d{2}:\d{2}[.,]\d+\s+(?P<level>...)`) tried first, then a
   bare word-boundaried severity token. On match, surfaces `level` with a ready-to-use two-stage
   `regexp` parse + filter hint.
4. Conservative throughout — nothing is fabricated; if no structure is detected, only what's
   actually matched in the sample is surfaced.

This unblocks runtime field-discovery for the log-only majority of the fleet that doesn't emit
JSON bodies.

## Test plan
- [x] `go build ./...` — success
- [x] `go vet ./...` — clean
- [x] `go test ./internal/telemetry/logs/...` — all pass, including new cases:
  - JSON body: existing behavior unchanged
  - logfmt body (`level=info msg="hi" status=200`): surfaces `level`, `status`, `msg` with logfmt hint
  - plaintext-inline (timestamp + INFO/ERROR token): surfaces `level` with a regexp hint whose
    pattern is verified to actually match a sampled ERROR line
  - unstructured line with no level token: no fields fabricated


internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -77,6 +77,74 @@ func bodyDerivedHint(key string) string {
...
+	return keys
+}
+
+// Well-known plaintext severity/level patterns, tried in priority order.
+// levelTimestampPattern anchors on a preceding HH:MM:SS(.mmm) timestamp
+// (common in Java/Log4j-style lines); levelBarePattern is the fallback
+// bare, word-boundaried severity token.
+const (
+	levelTimestampPattern = `\d{2}:\d{2}:\d{2}[.,]\d+\s+(?P<level>TRACE|DEBUG|INFO|WARN|WARNING|ERROR|FATAL)\b`
+	levelBarePattern      = `\b(?P<level>TRACE|DEBUG|INFO|WARN|WARNING|ERROR|FATAL)\b`
```
faiz9068: **Bare severity token → false positives; contradicts the "nothing fabricated" claim.** `\b(?P<level>...)\b` matches a severity word *anywhere* in the line:

- `100.65.18.112 GET /v1/INFO 200` → surfaces `level` from a URL path segment
- `no ERROR occurred during the run` → surfaces `level` from prose

Both invent a body-derived `level` for logs that have no structured level — across exactly the ~86% plaintext fleet this PR targets. The timestamp-anchored pattern above is safe; this bare fallback is the problem. Suggest dropping the bare fallback, or requiring the token near line-start / preceded by a timestamp or bracket.

internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -77,6 +77,74 @@ func bodyDerivedHint(key string) string {
...
+	}
+	return keys
+}
+
+// Well-known plaintext severity/level patterns, tried in priority order.
+// levelTimestampPattern anchors on a preceding HH:MM:SS(.mmm) timestamp
+// (common in Java/Log4j-style lines); levelBarePattern is the fallback
+// bare, word-boundaried severity token.
+const (
+	levelTimestampPattern = `\d{2}:\d{2}:\d{2}[.,]\d+\s+(?P<level>TRACE|DEBUG|INFO|WARN|WARNING|ERROR|FATAL)\b`
```
faiz9068: **Case-sensitive — misses the common lowercase form.** Tokens are uppercase-only, so `2026-07-08 12:00:00 error something` never matches. Lowercase `debug/info/warn/error` is very common in Go/Python/Node plaintext logs, so the anchored path silently misses a large real slice. Add `(?i)` (or lowercase alternates) to both `levelTimestampPattern` and `levelBarePattern`.

internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -95,48 +163,120 @@ func sampleBodyDerivedAttributes(ctx context.Context, client *http.Client, cfg m
...
+				freq[key]++
+			}
 			continue
 		}
-		for key := range obj {
-			if !safeBodyKeyPattern.MatchString(key) {
-				continue
+
+		// Not JSON: try logfmt (key=value pairs) next.
+		if kvKeys := parseLogfmtKV(line); len(kvKeys) >= 2 {
```
faiz9068: **`>= 2` is weaker than the doc comment implies — prose false positive.** `set a=1 where b=2 in prose` yields keys `a`, `b` → classified as logfmt → surfaces `a` and `b` as filterable Body fields. Consider also requiring the matched pairs to be a dominant fraction of the line's tokens (or anchoring) before declaring a line logfmt.

internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -95,48 +163,120 @@ func sampleBodyDerivedAttributes(ctx context.Context, client *http.Client, cfg m
...
 			}
-			freq[key]++
+			continue
+		}
+
+		// Not logfmt either: fall back to a well-known plaintext-inline
+		// severity token, preferring a timestamp-anchored match.
+		if levelTimestampRegexp.MatchString(line) {
+			levelTSCount++
+		} else if levelBareRegexp.MatchString(line) {
```
faiz9068: This `levelBareRegexp` branch is **never exercised by the tests** — `TestGetLogAttributesForPipeline_PlaintextInlineLevel` uses a full `HH:MM:SS.mmm` timestamp, so only the anchored branch (line 195) runs. The bare branch is both the untested path and the one with the false-positive surface (see comment on line 140). Please add a test that drives it (and, given the false positives, ideally one asserting a URL/prose line does *not* produce `level`).

internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -95,48 +163,120 @@ func sampleBodyDerivedAttributes(ctx context.Context, client *http.Client, cfg m
...
+	// token was found in at least one sampled line. Conservative — only
+	// surface "level" when the pattern actually matched a sampled line.
+	if levelTSCount > 0 || levelBareCount > 0 {
+		pattern, count := levelBarePattern, levelBareCount
+		if levelTSCount > 0 {
+			pattern, count = levelTimestampPattern, levelTSCount
+		}
+		return []LogAttribute{{
+			Name:           "level",
+			FilterField:    "attributes['level']",
```
faiz9068, line 341: **Duplicates an already-indexed severity for the exact fleet this targets.** This body-derived field gets `filter_field = attributes['level']`, while an ingest-populated stream severity comes back as `severity` → `SeverityText`. The merge dedups on exact `filter_field`, so these two keys differ and **both** are emitted. For a plaintext service whose `severity` is already set at ingest (which is what the dashboard actually renders — `getSanitizedSeverity(content.stream.severity)`, no client-side Body parse), the tool returns the real, directly-filterable field **plus** this inferior parallel `level` that tells the user to run a regexp parse they don't need. Suggest suppressing the body-derived `level` when an indexed `severity`/`SeverityText`/`level` is present — i.e. dedup the severity family by semantic name, not just exact `filter_field`.

internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -95,48 +163,120 @@ func sampleBodyDerivedAttributes(ctx context.Context, client *http.Client, cfg m
...
+			Hint:           regexpBodyDerivedHint(pattern, "level", "ERROR"),
```
faiz9068: **Hardcoded `"ERROR"` may not exist in the data.** Every other hint uses the `"<value>"` placeholder (`bodyDerivedHint`, `logfmtBodyDerivedHint`). Here a concrete `ERROR` is emitted even when the sample only contained INFO/DEBUG lines, so a user copying the hint verbatim filters for a value that may never occur. Use `"<value>"` for consistency.

faiz9068 (changes requested):
Reviewed the logfmt + plaintext-inline Body detection. Build/vet/tests pass and the structure is solid (package-level compiled regexes, hints via `json.Marshal`, JSON path unchanged and first). But two of the PR's headline claims don't hold under test, and the frontend premise cited in the description doesn't match the actual code. Details inline; summary:

**Conservativeness claim (#4) is empirically false.** The bare severity regex and the 2-pair logfmt heuristic both fabricate fields from prose/URLs (see inline on the pattern lines).

**Frontend premise is inaccurate.** The description says "the product UI (which does parse plaintext) shows them." It doesn't — the dashboard reads `severity` from backend stream metadata (`getSanitizedSeverity(content.stream.severity)` in `App/scenes/Logs/utils.tsx`); there is no client-side Body parsing for `level`. logfmt/regexp exist only as LogQL keywords executed backend-side, and the visual parse stage is JSON-only. So the justification for a body-derived `level` is weak, and it collides with an already-indexed `severity` (inline on the `level` construction).

**Cascade is result-level, not per-line** (`len(freq)>0` → `len(logfmtFreq)>0` → level): one stray JSON or logfmt line in an otherwise-plaintext sample suppresses all downstream detection. Worth a comment or per-line accumulation.

**Test gaps:** the bare-severity branch is never exercised (the one plaintext test uses a full timestamp, so only the anchored path runs); there is no dedup-vs-indexed-severity test; no coverage for logfmt-on-prose, lowercase levels, or mixed samples.

internal/telemetry/logs/attributes_for_pipeline.go
prathamesh-sonpatki: Fixed: dropped the bare mid-line `\bLEVEL\b` fallback. Replaced with a conservative `levelBracketOrStartPattern` that only matches severity at line-start or inside `[]`. URL/prose cases like `GET /v1/INFO` and `no ERROR occurred` no longer fabricate `level` (covered by `TestGetLogAttributesForPipeline_UnstructuredNoLevel`).

internal/telemetry/logs/attributes_for_pipeline.go
prathamesh-sonpatki: Fixed: both `levelTimestampPattern` and `levelBracketOrStartPattern` now use `(?i)`. Lowercase `error`/`info`/etc. match; covered by `TestGetLogAttributesForPipeline_PlaintextLowercaseLevel`.

internal/telemetry/logs/attributes_for_pipeline.go
prathamesh-sonpatki: Fixed: added `isLogfmtLine` — still requires ≥2 distinct KV pairs, plus matched pairs must cover ≥50% of the line's non-whitespace content. `set a=1 where b=2 in prose` is rejected (`TestGetLogAttributesForPipeline_LogfmtProseRejected`).

internal/telemetry/logs/attributes_for_pipeline.go
prathamesh-sonpatki: Fixed by removing the bare branch entirely (see sibling thread). Added `TestGetLogAttributesForPipeline_PlaintextBracketOrStartLevel` for the new conservative fallback, plus negative coverage for URL/prose mid-line severity words.

internal/telemetry/logs/attributes_for_pipeline.go
prathamesh-sonpatki: Fixed: merge now suppresses body-derived severity-family names (`level`/`severity`/`SeverityText`) when any indexed severity-family field is present — semantic dedup, not just exact `filter_field`. Covered by `TestGetLogAttributesForPipeline_SeverityFamilyDedup`.

internal/telemetry/logs/attributes_for_pipeline.go
prathamesh-sonpatki: Fixed: regexp hint now uses `"<value>"` like the JSON/logfmt hints.

faiz9068 (approved):

PR #184, opened 2026-07-13
Add bounded APM service deviation analysis

prathamesh-sonpatki:
## Summary
- add an APM service deviation tool for equal current and baseline RED comparisons
- report evidence quality, bounded classifications, telemetry changes, throughput shifts, and operation-level supporting evidence
- reconcile service Apdex movement with request-weighted operation contributions and an explicit unexplained residual
- preserve scope, window, result-limit, and response-redaction guarantees

## Test plan
- [x] `go test ./... -count=1`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] live MCP smoke test against an active service scope


internal/apm/deviations.go
```diff
@@ -0,0 +1,982 @@
...
+		left := math.Abs(result.ThroughputShifts[i].Comparison.AbsoluteDelta)
+		right := math.Abs(result.ThroughputShifts[j].Comparison.AbsoluteDelta)
+		if left != right {
+			return left > right
+		}
+		return identityLess(result.ThroughputShifts[i].ServiceName, result.ThroughputShifts[i].Env, result.ThroughputShifts[j].ServiceName, result.ThroughputShifts[j].Env)
+	})
+}
+
+func limitDeviationResult(result *apmDeviationResult, limit int) {
```
faiz9068, line 573: **Important — service cap is alphabetical, not severity-ranked, and can hide the worst deviation.**

`sortDeviationResult` orders `result.Services` by `identityLess` (alphabetical). `limitDeviationResult` then fills its identity budget by iterating services in that alphabetical order and filters the leaderboards down to those identities. The leaderboards are severity-sorted, but the identity cap that gates them is not.

Failure scenario: a fleet with 15 comparable services where the only regressions are in `z-api`/`z-worker`. With `max_services=10`, the budget is consumed by the 10 alphabetically-first (stable) services, so the `z-*` regressions get filtered out of the leaderboards entirely. Meanwhile `Outcome` was set to `deviations_detected` at line 279 *before* limiting — so the response claims deviations with empty leaderboards. For a "what regressed?" tool this hides exactly the signal the user asked for.

Only triggers when comparable services > `max_services` (fleet > 10), so all current tests (single-service/small-fleet) pass.

Fix: rank the identity budget by leaderboard presence/magnitude (severity) rather than alphabetically, and/or recompute `Outcome` after `limitDeviationResult` so it can't claim deviations the payload no longer contains.

internal/apm/deviations_stats.go
```diff
@@ -0,0 +1,215 @@
...
+func classifyEvidence(current, baseline MetricEvidence) EvidenceQuality {
+	if current.ObservedPoints == 0 || baseline.ObservedPoints == 0 {
+		return EvidenceQuality{Level: "non_comparable", Reasons: []string{"series_missing_in_one_or_both_windows"}}
+	}
+	if current.ObservedPoints < 4 || baseline.ObservedPoints < 4 {
+		return EvidenceQuality{Level: "sparse", Reasons: []string{"fewer_than_four_aligned_observations"}}
+	}
+	if current.ExpectedPoints <= 0 || baseline.ExpectedPoints <= 0 ||
+		current.ObservedPoints != current.ExpectedPoints || baseline.ObservedPoints != baseline.ExpectedPoints ||
+		current.ExpectedPoints != baseline.ExpectedPoints || current.ExcludedValues > 0 || baseline.ExcludedValues > 0 {
```
faiz9068, line 161: **Important — strict `observed == expected` coverage gate will under-report on real data (needs live verification).**

`classifyEvidence` returns `sparse` unless `ObservedPoints == ExpectedPoints` in both windows with zero exclusions, and `compareSignal` never ranks a `sparse`/`insufficient_evidence` signal. `ExpectedPoints` is `bucketCapacity = ceil(duration/step)`, but a single genuinely missing bucket (scrape gap, rollup timing, late data — common in production APM) drops the window below `expected` and silently downgrades the deviation to `insufficient_evidence`. The tool could stay silent during exactly the messy incidents it targets.

Also worth checking: with previous_period windows `baseline.End == current.Start`, and PromQL subqueries evaluate at both endpoints — verify the boundary bucket isn't double-counted / that `count_over_time` cardinality matches the `bucketCapacity` the code expects.

Part of this is by design (spec says "complete aligned coverage"), so confirm intent — but please verify against a live datasource before relying on it, and consider a small tolerance (e.g. coverage ≥ 0.95) rather than strict equality.

internal/apm/deviations_stats.go
```diff
@@ -0,0 +1,215 @@
...
+
+func selectedEvidence(signalName string, evidence WindowEvidence) MetricEvidence {
+	if evidence.Selected.ExpectedPoints != 0 || evidence.Selected.ObservedPoints != 0 ||
+		evidence.Selected.MissingValues != 0 || evidence.Selected.ExcludedValues != 0 {
+		return evidence.Selected
+	}
+	switch signalName {
+	case "request_rpm", "requests", "throughput":
+		return evidence.Requests
+	case "error_rpm", "errors":
```
faiz9068: **Minor** — the `selectedEvidence` fallback switch omits `"error_throughput_rpm"` (the name used in `deviationSignalDefinitions`); it only matches `"error_rpm"/"errors"`. In practice `signalSummary` always populates `Evidence.Selected`, so this fallback is dead for that signal — no bug today, but a latent inconsistency. Add the correct case name.

internal/apm/deviations.go
```diff
@@ -0,0 +1,982 @@
...
+		return
+	}
+	target[name] = Distribution{Q25: *q25, Median: *median, Q75: *q75, IQR: *q75 - *q25}
+}
+
+func aggregateCount(value *float64, expected int) int {
+	if value == nil || !isFinite(*value) || *value <= 0 {
+		return 0
+	}
+	return min(int(math.Round(*value)), expected)
```
faiz9068, line 444: **Minor** — because observed count is capped at `expected` here, a window whose subquery returns one extra boundary sample while also missing one interior bucket nets to `expected` and reports coverage 1.0 despite a real gap. Low impact given the equality requirement in `classifyEvidence`, but worth a comment so the masking behavior is explicit.

internal/apm/deviations_window.go
```diff
@@ -0,0 +1,171 @@
...
+		return DeviationWindows{}, fmt.Errorf("current and baseline effective windows must contain equal completed buckets")
+	}
+
+	return DeviationWindows{
+		TimeSource:             timeSource,
+		BaselineMode:           baselineMode,
+		QueryStepSeconds:       int64(queryStep / time.Second),
+		RequestedCurrentStart:  requestedCurrent.Start,
+		RequestedCurrentEnd:    requestedCurrent.End,
+		RequestedBaselineStart: requestedBaseline.Start,
```
faiz9068, line 108: **Minor** — for the default/previous_period path, reported `RequestedBaselineStart/End` is anchored to the *requested* current start (with sub-minute precision), while the effective baseline is anchored to the *effective* current start. Provenance display only, but a model reading `requested_*` alongside `effective_*` may find them mildly inconsistent.

internal/apm/deviations.go
```diff
@@ -0,0 +1,982 @@
...
+	return leftService+"\x00"+leftEnv < rightService+"\x00"+rightEnv
+}
+
+func hasMaterialDeviation(result apmDeviationResult) bool {
+	return len(result.Leaderboards.Reliability.Regressions)+len(result.Leaderboards.Reliability.Improvements)+
+		len(result.Leaderboards.Experience.Regressions)+len(result.Leaderboards.Experience.Improvements)+
+		len(result.Leaderboards.SustainedLatency.Regressions)+len(result.Leaderboards.SustainedLatency.Improvements)+len(result.ThroughputShifts) > 0
+}
+
+func shouldQueryOperations(result apmDeviationResult) bool {
```
faiz9068, line 659: **Minor** — operation correlation and Apdex reconciliation only run when a *regression* exists, so pure-improvement service results get no operation breakdown. Likely intentional (investigation is regression-driven), but currently undocumented — a one-line comment here would help.

faiz9068 (commented):
Code review of the APM service deviations feature. Verified via `go build`/`go vet`/`go test ./internal/apm/... -count=1` (all pass, incl. `-race`) and hand-checking the math. No Critical issues. Two Important findings below (both only manifest on the fleet path >10 services, which is why CI is green), plus minor cleanups. Line comments follow.

internal/apm/deviations.go
faiz9068: **Correction — retracting this finding.** After tracing the query layer, this can't happen. `buildDeviationCandidateMask` (deviations_query.go:186-196) builds a single `topk(maxServices, ((currentTotal + baselineTotal) or currentTotal or baselineTotal))` mask and applies it to **both** the current and baseline window queries (lines 181-182). So the identity universe is capped to ≤ `maxServices` (ranked by combined request volume, not alphabetically) *before* any in-process code runs.

Since `|Services| + |telemetry_changes| = |union| ≤ maxServices`, `limitDeviationResult`'s identity budget always fits everything — leaderboards are never truncated and `Outcome` cannot desync. `limitDeviationResult` is effectively a defensive no-op here. My original analysis missed the candidate-mask semantics. No change needed.

internal/apm/deviations_stats.go
faiz9068: **Reframing — this is by-design, not a bug.** The strict `observed == expected` gate is the "complete aligned coverage" requirement from the evidence/safety contract (ENG-1463), so relaxing it to a tolerance is a product/contract decision rather than a fix.

The one genuinely actionable part is verification, not code: confirm against a **live datasource** that (a) `count_over_time` on the subqueries yields the `bucketCapacity` the code expects, and (b) the boundary bucket (previous_period `baseline.End == current.Start`, subquery endpoint inclusion) isn't double-counted. If live data shows the gate silently suppresses known regressions, revisit the tolerance *as a contract change*. No code change proposed here.

internal/apm/deviations_stats.go
faiz9068: Fixed in 50ab4b5 — added `error_throughput_rpm` to the fallback case. Note this switch is currently unreachable (`signalSummary` always populates `Evidence.Selected`), so it's correctness hygiene rather than a live fix.

internal/apm/deviations.go
faiz9068: Documented in 50ab4b5 — added a comment noting the cap keeps the coverage gate well-defined and that an extra boundary sample can net-cancel a missing interior bucket. Logic unchanged (the cap is relied on by the strict-equality gate); the live-data verification remains the actionable follow-up.

internal/apm/deviations.go
faiz9068: Documented in 50ab4b5 — added a comment on `shouldQueryOperations` clarifying that operation correlation/reconciliation is intentionally regression-driven, so pure-improvement results get no operation breakdown.

internal/apm/deviations_window.go
faiz9068: **Not changing this — it's intended behavior, not an inconsistency.** The `requested_*` fields are meant to record exactly what the caller asked for; the `effective_*` fields record the step-aligned window actually measured. For a default/relative window `requested_current_start` is `now - duration` (sub-minute), so the previous_period `requested_baseline_*` derived from it is sub-minute too, while the effective baseline is anchored to the step-aligned effective current start. That divergence is the whole point of the requested-vs-effective split in `DeviationWindows` — collapsing them would erase the provenance signal that tells a reader how much the window was trimmed. Leaving as-is.

faiz9068 (approved):

PR #185, opened 2026-07-13
feat(dashboards): add MCP tools for dashboard snapshots

prathamesh-sonpatki:
## Summary
- Add `list_dashboard_snapshots`, `get_dashboard_snapshot`, and `delete_dashboard_snapshot` MCP tools against the Last9 v4 `/dashboards/snapshots` API (ENG-1133).
- Preserve upstream `dashboard not found` 404s (kept distinct from `snapshot not found`), map 403s to an access-neutral message, and cap oversized success bodies at 5 MiB.
- Document tools in README and embed descriptions via `go:embed`.

## Test plan
- [x] `go test ./internal/dashboards/ ./...`
- [x] Integration: snapshot list/get/delete e2e (seeds a snapshot via the v4 API), gitignored `.env` / `TEST_REFRESH_TOKEN` (or `LAST9_REFRESH_TOKEN`)
- [x] `go run . dump-tools` shows the three snapshot tools


prathamesh-sonpatki:
## E2E (local, gitignored `.env`)

Ran against `app.last9.io` org `last9` using `LAST9_REFRESH_TOKEN` + `LAST9_DELETE_TOKEN`:

- `TestDashboardSnapshotCRUD_Integration` — **PASS** (create → list → get → delete)
- Root cause found during probe: omitting `variables` makes v4 create return 500; MCP now defaults to `{}` (commit `93cfffc`).

internal/dashboards/http.go
```diff
@@ -14,6 +14,10 @@ import (
 
 const maxAPIErrorBodyBytes = 4096
 
+// maxAPISuccessBodyBytes aligns with the monorepo dashboard snapshot create
```
faiz9068: **Should fix — this guard is asymmetric with the backend.** The backend caps the create *request* at 25 MiB (`http.MaxBytesReader` in `createDashboardSnapshotHandler`). This caps the *response*. But `create` echoes back the full snapshot = the same blobs **plus** envelope fields (`id`, `organization_id`, `created_by`, `created_at`), so the response is strictly larger than the request.

Consequence:
- A snapshot within ~200 bytes of the cap is **persisted by the backend but reported as failed** here (line 50) → caller may retry → duplicate snapshots.
- `get_dashboard_snapshot` on such a snapshot exceeds the guard on **every** read → the valid, stored snapshot is **permanently unreadable** via MCP.

Low probability (only near-25 MiB payloads) but bad failure mode. Suggest giving the response guard headroom above the backend's request cap (e.g. `32 * 1024 * 1024`) and updating this comment — the two limits measure different payloads, so they should intentionally *not* be equal.

internal/constants/api.go
```diff
@@ -39,8 +39,10 @@ const (
 	EndpointSuggest = "/suggest"
 
 	// Dashboard API endpoints (v4)
-	EndpointDashboards    = "/dashboards"
-	EndpointDashboardByID = "/dashboards/%s" // fmt with id; GET requires ?region=
+	EndpointDashboards           = "/dashboards"
```
faiz9068: **gofmt nit.** The new longer `EndpointDashboardSnapshotByID` changes the const-block alignment; `gofmt -w` re-aligns the `=` on these four lines. `gofmt -l` currently flags this file.

internal/dashboards/get_snapshot_test.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func TestGetDashboardSnapshotHandler_GETsByID(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet || r.URL.Path != "/dashboards/snapshots/snap-1" {
+			http.NotFound(w, r)
+			return
+		}
+		_ = json.NewEncoder(w).Encode(map[string]any{
+			"id":                    "snap-1",
```
faiz9068: **gofmt nit.** These map keys are over-padded by one space; `gofmt -w` fixes the alignment. `gofmt -l` currently flags this file.

faiz9068 (commented):
End-to-end contract + code review. Verified the wire format against the merged backend (`last9/last9#11143`: `handler_dashboard_snapshots.go`, `spec/dashboard_snapshots.go`, `store/dashboard_snapshots.go`, `routes_v4.go`) and the frontend snapshot route. Build + unit tests green locally; head CI green.

**Verified correct:** request/response fields, endpoints/methods, `dashboard_id` query param and `{snapshot_id}` path all match `spec.CreateDashboardSnapshotRequest`/`routes_v4.go` 1:1; deeplink matches the frontend `buildSnapshotUrl`; the `variables={}` workaround is genuinely required (nil `json.RawMessage` → empty string → invalid for the `variables JSONB` column → 500); error disambiguation is sound (`"dashboard snapshot not found"` doesn't contain the substring `"dashboard not found"`); the hand-written create InputSchema is the right call for the `json.RawMessage` fields.

**Verdict: ready to merge with fixes** — one should-fix (response-guard asymmetry) + two gofmt nits, all inline below. Everything load-bearing is verified correct.

internal/constants/api.go
faiz9068: ✅ Fixed in b54c61d (`gofmt -w`).

internal/dashboards/get_snapshot_test.go
faiz9068: ✅ Fixed in b54c61d (`gofmt -w`).

internal/dashboards/http.go
faiz9068: Addressed across the follow-up commits:

- `create_dashboard_snapshot` was **dropped from this PR** (server-side capture pending — ENG-1492), so the create-echo half of this finding (create succeeds on the backend but this reports failure → duplicate retries) no longer applies.
- The cap now targets `get_dashboard_snapshot` and was lowered **25 MiB → 5 MiB** (2993537). At 5 MiB a large `get` returns the "open the UI link" error instead of dumping MBs into the client — that's the intended behavior for a read tool, not a false failure. Comment reworded to match; no longer claims parity with a backend request cap.

Resolving.

faiz9068 (approved):
.

PR #186, opened 2026-07-15
Add bounded trace investigation tools

prathamesh-sonpatki:
## Summary

Adds MCP-facing trace investigation tools:

- `get_trace_attribute_deviations` returns bounded, ranked cohort evidence
- `get_trace_waterfall` returns exact-trace timing and parent/child evidence
- trace-detail decoding includes the fields required by these workflows
- versioned contracts, fixtures, feature metadata, and compatibility tests are included
- tool descriptions remain embedded Markdown with generated schemas

## Behavior

The tools return bounded observations and correlation evidence without making causal or critical-path claims. Safety, cardinality, and response limits remain authoritative at the data boundary.

## Validation

- full test suite passes: 526 tests across 16 packages
- focused trace tests pass: 194 tests
- static analysis passes
- generated tool schemas contain both tools and reject unknown arguments
- contracts, fixtures, and formatting checks pass

## Dependency

The deviation tool requires the companion default-off backend capability to land and be enabled first.


tools.go
```diff
@@ -229,6 +229,13 @@ func registerAllTools(server *last9mcp.Last9MCPServer, cfg models.Config) error
 		Description: prompts.GetTraceAttributeValuesDescription,
 	}, traces.NewGetTraceAttributeValuesHandler(client, cfg)))
 
+	last9mcp.RegisterInstrumentedTool(server, &mcp.Tool{
```
faiz9068: **Registration bypasses `registerIfAllowed`.**

Every other tool in this file goes through `reg(registerIfAllowed(server, cfg.AllowedTools, ...))`. These two call `last9mcp.RegisterInstrumentedTool` directly, which has three effects:

1. Toolset filtering is skipped. Verified with the built binary: `dump-tools --toolsets=dashboards` returns 10 tools **including** `get_trace_attribute_deviations` and `get_trace_waterfall`. That breaks the `Set.Allows` hard-filter contract, and `CONTEXT.md` treats `dump-tools` under toolset selection as canonical.
2. The returned `error` is discarded, so a registration failure is silent instead of failing startup.
3. No `recover()` — and the comment on `registerIfAllowed` says the SDK *panics* on invalid tool schemas, so a bad schema now crashes at boot rather than returning a clean error.

Neither tool is in `internal/toolsets/toolsets.go` either. Suggest wrapping both in `reg(registerIfAllowed(...))` and adding them to the `traces` membership list.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,332 @@
...
+		dl := deeplink.NewBuilder(cfg.OrgSlug, cfg.ClusterID).BuildTracesLink(start.UnixMilli(), end.UnixMilli(), nil, args.TraceID, "")
+		return &mcp.CallToolResult{Meta: deeplink.ToMeta(dl), Content: []mcp.Content{&mcp.TextContent{Text: string(b)}}}, nil, nil
+	}
+}
+
+type spanInterval struct{ start, end int64 }
+
+func buildTraceWaterfall(traceID string, raw []TraceDetailsSpan, limit int, selectedID string) TraceWaterfallResponse {
+	var resp TraceWaterfallResponse
+	resp.ContractVersion = "investigation-evidence/v1"
```
faiz9068: **This stamps a contract version the payload does not satisfy.**

Validating a real `buildTraceWaterfall` output against this PR's own `contracts/investigation-evidence-v1.schema.json`:

```
/$defs/evidence: required: missing properties: ["partial" "provenance"]
```

Past that first failure, `TraceWaterfallResponse` also lacks the required top-level `analysis_version` and `request` (with `scope`, `requested_window`, `effective_window`), and `interpretation` has no `claim_type`.

`contracts_test.go` only validates the hand-written fixtures, never generated output, which is why CI is green. So `contracts/README.md`'s rule "Producers emit exact major versions" isn't true for the only producer in the repo, and a consumer written against the schema would reject live tool output.

Either emit the full envelope (`analysis_version: "trace-waterfall/v1"`, `request`, `evidence.partial`, `evidence.provenance`, `interpretation.claim_type`) and add a test that validates *marshalled* output against the schema, or drop `contract_version` until it conforms.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,332 @@
...
+	resp.TraceID = traceID
+	resp.Evidence.AppliedLimit = limit
+	resp.Evidence.Truncated = len(raw) >= limit
```
faiz9068: **Truncation is judged after the environment filter, so it under-reports.**

`raw` here is the post-`traceDetailsMatchesEnv` slice, but the limit was applied by the API to the pre-filter set. If the API returns exactly `limit` spans (i.e. truncated) and env filtering drops any of them, this reports `truncated: false` on a partial waterfall.

That matters beyond the flag: when children are cut, the parent's `self_time_ms` is silently inflated. On a real 500-span trace the root reported `self_time_ms: 183214` of `duration_ms: 184526` — nearly all of it attributed as self time because its children were truncated away, with `warnings: []`.

Suggest capturing the pre-filter count and testing against that, plus a warning when truncation makes self-time unreliable.

faiz9068 (commented):
Reviewed the served behaviour by building the binary and calling both tools over stdio against a live backend, so most of these are verified rather than inferred. Grouped roughly by severity in the inline comments.

Two blockers in my view: the registration path bypassing `registerIfAllowed` (both tools leak into every toolset), and `get_trace_waterfall` stamping `investigation-evidence/v1` on a payload that fails that schema.

`get_trace_waterfall` itself computes correctly on real data — interval-union self-time held up on a 500-span trace (no span with self time exceeding its duration, none outside the reported window), `selected_span_id` round-trips, and `max_spans` bounds validate. The issues below are about what the payload *claims* versus what it contains.

Not inlined because the files aren't in the diff: `README.md` has no entry for either tool, while every other tool has both a list bullet and a section.

The description-token-budget removal is intentional per the author, so I've left it out.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	resp.Interpretation.Summary = "Compact trace waterfall with interval-union self-time. Slowest spans and self-time contributors are observations, not proof of cause."
+	resp.Interpretation.Limitations = []string{"V1 does not compute or claim a critical path."}
+	if resp.Spans == nil {
+		resp.Spans = []WaterfallSpan{}
+	}
+	return resp
+}
+
+func parseRFCNano(v string) int64 {
```
faiz9068: **`0` on parse failure is then used as a real instant.**

`parseRFCNano` returns `0` when the timestamp doesn't parse, and `minStart == 0` (line 181) doubles as the "unset" sentinel. With one unparseable timestamp among two spans, the output was:

```json
"summary": { "duration_ms": 50 },          // should be 100
"spans": [{ "start_offset_ms": -1784073600009.9998, "self_time_ms": 100 }]  // should be 50
```

plus a spurious `child interval outside parent` warning. So a nonsensical offset, a wrong trace duration, and a wrong self time are handed to the model as timing evidence at `evidence_quality: medium`.

Suggest `func parseRFCNano(v string) (int64, bool)`, excluding the span with an explicit warning on failure, and a separate `haveBounds bool` instead of the `minStart == 0` sentinel. Note `parseTimestampToUnix` in `service_traces.go` has the same two-format limitation — one nanosecond-precision helper would fix both.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,332 @@
...
+	resp.Summary.ServiceCount = len(services)
+	resp.Summary.RootSpanIDs = roots
+	resp.Summary.DurationMs = float64(maxEnd-minStart) / 1e6
+	if minStart > 0 {
+		resp.Summary.Start = time.Unix(0, minStart).UTC().Format(time.RFC3339Nano)
+		resp.Summary.End = time.Unix(0, maxEnd).UTC().Format(time.RFC3339Nano)
+	}
+	resp.Evidence.ReturnedSpans = len(resp.Spans)
+	resp.Evidence.Warnings = warnings
+	resp.Interpretation.EvidenceQuality = "high"
```
faiz9068: **Empty results are reported as high-quality evidence.**

Verified against a nonexistent trace ID: `span_count: 0`, `warnings: []`, `evidence_quality: "high"`. Nothing found should not be high confidence — a "trace not found in window" warning and a downgrade to `insufficient` would keep the model from treating an empty waterfall as a negative result.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		if _, ok := byID[s.SpanID]; ok {
+			warnings = append(warnings, "duplicate span ID: "+s.SpanID)
+			continue
+		}
+		byID[s.SpanID] = s
+	}
+	children := map[string][]string{}
+	roots := []string{}
+	for id, s := range byID {
```
faiz9068: **Warning order is non-deterministic.**

This loop and the one at line 225 both iterate `byID`, a Go map, appending to `warnings`. Measured 6 distinct `Evidence.Warnings` orderings across 50 runs on identical input.

Given the PR's deterministic-fixture theme and that tool output is pinned via `dump-tools`/eval snapshots, this will produce flaky diffs — and it makes the SHA-256 evidence-immutability rule in `contracts/README.md` unimplementable for this producer. `sort.Strings(warnings)` before assigning is enough.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,332 @@
...
+				intervals = append(intervals, spanInterval{cs, ce})
+			}
+		}
+		self := s.Duration - unionDuration(intervals)
+		if self < 0 {
+			self = 0
+		}
+		w := WaterfallSpan{SpanID: id, ParentSpanID: s.ParentSpanID, Service: s.ServiceName, Operation: s.SpanName, Kind: s.SpanKind, Status: s.StatusCode, StartOffsetMs: float64(st-minStart) / 1e6, DurationMs: float64(s.Duration) / 1e6, SelfTimeMs: float64(self) / 1e6, Depth: depth[id]}
+		resp.Spans = append(resp.Spans, w)
+		services[s.ServiceName] = true
```
faiz9068: Minor: an empty `ServiceName` is counted as a service here, inflating `summary.service_count` when the field is missing.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,332 @@
...
+		Limitations     []string `json:"limitations"`
+	} `json:"interpretation"`
+}
+
+func NewGetTraceWaterfallHandler(client *http.Client, cfg models.Config) func(context.Context, *mcp.CallToolRequest, GetTraceWaterfallArgs) (*mcp.CallToolResult, any, error) {
+	return func(ctx context.Context, _ *mcp.CallToolRequest, args GetTraceWaterfallArgs) (*mcp.CallToolResult, any, error) {
+		if args.TraceID == "" {
+			return nil, nil, fmt.Errorf("trace_id is required")
+		}
+		max := args.MaxSpans
```
faiz9068: Minor: `max` shadows the builtin (legal since Go 1.21, but the repo generally names these `limit`/`maxSpans`).

internal/telemetry/traces/deviations.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil, nil
+}
+
+func newDeviationAPIRequest(args GetTraceAttributeDeviationsArgs, mode string, target, control deviationAPIWindow, threshold *float64) deviationAPIRequest {
+	return deviationAPIRequest{
+		ContractVersion: attributeDeviationsVersion,
+		Scope:           deviationAPIScope{ServiceName: args.ServiceName, Environment: args.Environment, Operation: args.Operation, Filters: args.Filters},
+		Comparison:      deviationAPIComparison{Mode: mode, Target: target, Control: control, LatencyThresholdMs: threshold},
+		Candidates:      deviationAPICandidates{Attributes: args.CandidateAttributes, AutoDiscover: len(args.CandidateAttributes) == 0},
+		Limits:          deviationAPILimits{MinimumCohortSize: args.MinimumCohortSize, MinimumValueSupport: args.MinimumValueSupport, MaximumCandidates: len(args.CandidateAttributes), MaximumRankedResults: args.Limit},
```
faiz9068: **The documented limits are never applied, and three fields are always zero.**

The description promises `candidate_attributes` max 8, `limit` default 10 / max 10, `minimum_cohort_size` default 100 / min 20, `minimum_value_support` default 20 / min 10. None are validated or defaulted here — raw zero values are forwarded. `MaximumValuesPerAttribute` and `RepresentativesPerResult` are never assigned at all, and `MaximumCandidates: len(args.CandidateAttributes)` is `0` on the auto-discover path, so the request asks for "maximum 0 candidates".

This is observable against the live endpoint: every successful call returns `warnings: ["candidate_limit_reached"]`, i.e. the backend substitutes its own bound and every call gets a truncated candidate set that the caller never asked for. Passing 99 candidate attributes and `limit: 999` also reaches the network unclamped.

Suggest applying the documented defaults and clamps in `buildDeviationAPIRequest`, rejecting more than 8 candidates, and either populating or deleting the two unused limit fields.

internal/telemetry/traces/deviations.go
```diff
@@ -0,0 +1,221 @@
...
+	response, err := client.Do(request)
+	if err != nil {
+		return nil, fmt.Errorf("attribute deviations API request failed: %w", err)
+	}
+	defer response.Body.Close()
+	responseBody, readErr := io.ReadAll(io.LimitReader(response.Body, 4<<20))
+	if readErr != nil {
+		return nil, readErr
+	}
+	if response.StatusCode != http.StatusOK {
```
faiz9068: **Upstream response body is echoed for every non-200, bypassing the sanitization added in #188.**

`internal/telemetry/traces/errors.go` (merged just before this PR's base) established the pattern: `newTraceHTTPErrorWithHint` echoes the upstream body **only** for 400/422, caps it, drains it otherwise, and attaches a validated request ID. This formats the raw body into the error for every non-200 — 500, 502, 504 included — and `deviations_test.go` asserts that a 504 body leaks through.

A query timeout body is exactly where backend query text tends to appear, so this is the case the sanitization layer exists for. Suggest using `newTraceTransportError` / `newTraceHTTPError` here and inverting that test to assert the body is *not* echoed for 5xx.

Related: the status check in `waterfall.go` returns a bare `status %d` and skips `newTraceNotFoundTraceError`, so callers lose the 404 "widen the window" hint.

contracts_test.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+		seen[name] = true
+	}
+	if len(seen) != 7 {
+		t.Fatalf("got %d flags, want 7", len(seen))
+	}
+}
+
+func TestInvestigationVersionCompatibility(t *testing.T) {
+	acceptsEvidence := func(version string) bool { return version == evidenceV1 }
```
faiz9068: **This test is a tautology.**

`acceptsEvidence` and `acceptsWorkflow` are closures defined here, and the assertions below check that those closures accept `v1` and reject `v2` — testing a literal written three lines above, not any production code path. The "additive optional field" half sorts a key slice, discards it, then asserts `contract_version` is unchanged, which it cannot be.

Either point the accept/reject logic at whatever function will actually validate incoming contract versions, or drop the test rather than bank confidence on it.

internal/prompts/descriptions/get_trace_attribute_deviations.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Parameters
+
+- `comparison_mode` (required): `latency`, `errors`, or `time`.
+- `service_name` (required): exact service name.
+- `environment` (required): exact `deployment.environment` value.
+- `operation`: optional exact operation/span name.
+- `filters`: optional trace JSON filter conditions. Discover valid fields first with `get_trace_attributes_for_pipeline`.
+- `candidate_attributes`: up to 8 raw attribute names or returned `filter_field` values. Omit for bounded safe discovery. Sensitive and identifier-like fields are always excluded.
+- `latency_threshold_ms`: required for latency mode. The tool converts milliseconds to the trace API's nanosecond `Duration` field.
```
faiz9068: This is inaccurate: the tool forwards `latency_threshold_ms` verbatim in the request payload and the conversion happens server-side. Given the repo's own note that unit mistakes in descriptions propagate into model behaviour, worth correcting or dropping.

Also on this file — "computed in one ClickHouse statement" is an implementation detail the model can't act on.

internal/prompts/descriptions/get_trace_waterfall.md
```diff
@@ -0,0 +1,13 @@
...
+
+Parameters:
+- `trace_id` (required): exact trace ID.
+- `environment`: optional exact deployment environment.
+- `start_time_iso` / `end_time_iso`: optional RFC3339 bounds.
+- `lookback_minutes`: default 4320 for exact lookup.
+- `selected_span_id`: include attributes, events, and links for this span only.
+- `max_spans`: default 500, maximum 1000.
+
+Self-time subtracts the union of clipped direct-child intervals. The tool does not compute a critical path and does not prove root cause. Treat truncation, cycles, duplicates, missing parents, and clock-skew warnings as evidence limitations.
```
faiz9068: "clock-skew warnings" doesn't match what's emitted — the actual string is `child interval outside parent`. Aligning the wording lets the model reason about warnings it will really receive.

tools.go
faiz9068: Fixed in efe621d. Both tools now go through `reg(registerIfAllowed(...))` and are listed under `traces` in `internal/toolsets/toolsets.go`, which also places them in the `investigate` composite.

Verified with a fresh binary: `dump-tools --toolsets=dashboards` went from 10 tools to 8 with neither tool present; `--toolsets=traces` (7) and `--toolsets=investigate` (29) both include them.

Also updated the AGENTS.md registration snippet, since it documented the bare `RegisterInstrumentedTool` call and is what the next tool would copy.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 41a4b77 by emitting the full envelope rather than dropping the version: `analysis_version`, `request{scope,requested_window,effective_window}`, `evidence{partial,provenance}`, `interpretation.claim_type`, with the waterfall payload under `data`.

Added `waterfall_contract_test.go`, which validates **marshalled** output against `contracts/investigation-evidence-v1.schema.json` on three paths (populated, empty, truncated+degraded). That is the gap that kept CI green.

Two things worth calling out. `request.scope` requires `service_name` and `environment` with `minLength: 1`, which a trace-ID lookup does not have — `service_name` is derived from the root span and `environment` is `unspecified` when the argument is omitted, both documented in the tool description. And this changes the tool output shape, so any pinned snapshot needs regenerating.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 41a4b77. Truncation is judged on the pre-filter fetched count.

The self-time consequence you flagged is handled too: a truncated result now emits a warning that `self_time_ms` may be overstated and adds a matching entry to `interpretation.limitations`, so the inflated root self time is no longer reported with empty warnings. Covered by `TestBuildTraceWaterfallTruncationJudgedBeforeEnvironmentFilter`.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 41a4b77, taking the suggested shape: `parseTraceTimestampNano(v) (int64, bool)`. Spans with an unparseable timestamp are excluded with an explicit warning, and a separate `haveBounds` replaces the `minStart == 0` sentinel.

Consolidated with `parseTimestampToUnix` as you suggested — it now delegates to the same parser, which also picked up the zone-less and space-separated layouts it was missing. Tests cover both the exclusion and epoch-zero as a legal instant.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 41a4b77. An empty result is now `evidence_quality: insufficient` with `claim_type: observation`, a `no spans found for this trace_id in the requested window` warning, and a limitation stating an empty waterfall is not evidence the trace does not exist.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 41a4b77. `sort.Strings(warnings)` before assignment, plus sorted iteration in the graph passes rather than ranging the map directly.

Two tests pin it: one asserts identical warning ordering across 50 runs, the other asserts the marshalled payload is byte-identical across 50 runs, so the content-hash rule is implementable for this producer.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 41a4b77 — an empty `ServiceName` is no longer counted, with a test.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 41a4b77 — renamed to `maxSpans`.

internal/telemetry/traces/deviations.go
faiz9068: Fixed in 1a38701. The documented defaults are applied and out-of-range values are rejected with a message naming the bound, rather than clamped — a caller that asked for `limit: 999` should be told, not silently corrected. More than 8 candidates is rejected too.

`maximum_candidates` no longer goes out as 0 on the auto-discover path, and both previously-unassigned limit fields are now populated at the bounds the endpoint applies by default. I checked the emitted payloads against the endpoint request validation: all shapes are accepted and no value is substituted, which is the condition this comment was about.

On the `candidate_limit_reached` observation — that warning fires whenever discovery fills its candidate ceiling, so it will still appear on most discovery calls and is not itself a sign of a malformed request. The tool description now says so and points at passing explicit `candidate_attributes` to look wider.

internal/telemetry/traces/deviations.go
faiz9068: Fixed in 1a38701. The status is checked before the body is read, so `newTraceHTTPError` owns it: 400/422 relayed sanitized and capped, everything else drained.

Inverted the 504 test to assert the body is **not** echoed, and added a 400 test asserting a caller-fixable body still is, so both halves of the sanitization contract are pinned.

Also fixed the related point: the waterfall status check now returns `newTraceNotFoundTraceError()` on 404 so the widen-the-window hint is preserved, `newTraceHTTPError` on other non-200, and the sanitized transport/invalid-response errors otherwise.

contracts_test.go
faiz9068: Fixed in 6123c60. Replaced with assertions against the schemas themselves rather than a locally defined predicate: the version `const` rejects an unknown major version, and the open `additionalProperties` accepts an unknown optional field. Applied to both the evidence and workflow schemas.

internal/prompts/descriptions/get_trace_attribute_deviations.md
faiz9068: Both fixed in 2b5ef7a. The threshold is forwarded as milliseconds and the split is applied server side, so the description now states that and notes it must be positive and omitted for the other modes. Dropped the storage-engine sentence.

internal/prompts/descriptions/get_trace_waterfall.md
faiz9068: Fixed in 2b5ef7a. The description now lists the warning strings the tool actually emits, including `child interval outside parent`, and explains what a truncated result means for `self_time_ms` and what `insufficient` evidence quality means.

internal/telemetry/traces/waterfall.go
faiz9068: Follow-up in c5669f4: the synthesized scope is gone.

I flagged above that `request.scope` forced this tool to manufacture a `service_name` and an `environment` it was never given. That was the wrong place to solve it — the schema assumed all evidence is service and environment scoped, which is false for trace-scoped evidence. `request.scope` now accepts either that form or a `trace_id` form, and the waterfall reports the trace ID it was actually given plus `environment` only when the caller filtered by one. Services stay readable from `data.spans` and `data.summary.service_count`.

The relaxation is strictly more permissive — the first branch is the previous `required` list — so every payload that validated before still validates. A test pins it so it is not vacuous: a scope with neither form, an empty `trace_id`, and a `service_name` without an `environment` are all rejected. The waterfall fixture now shows the trace-scoped form and its pinned content hash was recomputed.

dump_test.go
```diff
@@ -143,70 +143,6 @@ func TestDumpTools(t *testing.T) {
...
 
-func descriptionTokenEstimate(descs []string) int {
-	total := 0
-	for _, d := range descs {
-		total += len(d)
-	}
-	return total / 4 // chars/4 regression heuristic — not an exact tokenizer count
-}
-
-func TestDescriptionTokenBudgets(t *testing.T) {
```
faiz9068, line 154: **Important — the description-token-budget gate was removed because this PR's descriptions overrun it.**

Measured on both revisions with the estimator this test used (chars/4):

| surface | base `740df9f3` | head `a8558515` | deleted ceiling |
|---|---|---|---|
| `all` | 11,609 | **12,915** | 12,000 |
| `investigate` | 9,173 | **10,478** | 10,000 |

The two new description files add 5,232 chars (`get_trace_attribute_deviations.md` 3,034, `get_trace_waterfall.md` 2,198), which is what crosses both ceilings. The commit trail reads trim (`22d6539`) -> restore (`f246af5`) -> drop the test (`fdeacff`), and the same change also removes the gate line from `AGENTS.md` and the decision block from `CONTEXT.md`.

Why it matters: this was the only regression tripwire on served-surface token cost, in a repo whose `CONTEXT.md` is largely about that cost. Whale tripwires and toolset filtering bound *per-tool* size; they do not bound the aggregate surface, so nothing replaces what was deleted.

Not asking for a specific outcome — it needs an owner decision. Either trim the two descriptions back under 12k/10k and restore the gate, or raise the ceilings with the measured numbers and the rationale recorded in `CONTEXT.md`. A silent deletion leaves the next description-heavy PR with nothing to fail against.

internal/telemetry/traces/deviations.go
```diff
@@ -0,0 +1,268 @@
...
+	defer response.Body.Close()
+	// Status before body: newTraceHTTPError echoes only 400/422 (sanitized), drains 5xx.
+	if response.StatusCode != http.StatusOK {
+		return nil, newTraceHTTPError(response)
+	}
+	responseBody, readErr := io.ReadAll(io.LimitReader(response.Body, 4<<20))
+	if readErr != nil {
+		return nil, newTraceInvalidResponseError(readErr)
+	}
+	if !json.Valid(responseBody) {
```
faiz9068, line 271: **Important — the upstream payload is passed through with no contract check.**

`json.Valid` is the only validation here, so the response is forwarded whatever its shape. But the embedded description tells the model it is receiving an `investigation-evidence/v1` envelope with `evidence.warnings`, full-denominator shares, deterministic ranks and evidence quality. Nothing in this repo asserts that: the deviations fixture is hand-written and never compared against producer output, which is exactly the asymmetry the waterfall's contract test was added to close.

Failure mode: if the endpoint ships a different shape — or a partially-enabled path answers 200 with something else — the model is told it received conforming evidence and reasons over fields that are not there.

Minimum fix: decode enough to check `contract_version == "investigation-evidence/v1"` and fail with a clear error (or attach a warning) when it does not match. Better: an `httptest` server returning the pinned fixture, asserting the handler's output validates against the evidence schema, mirroring `waterfall_contract_test.go`.

internal/telemetry/traces/deviations.go
```diff
@@ -0,0 +1,268 @@
...
+	return deviationAPIWindow{Start: start, End: end}, nil
+}
+
+func deviationControlWindow(args GetTraceAttributeDeviationsArgs, mode string, target deviationAPIWindow) (deviationAPIWindow, error) {
+	if mode != "time" {
+		return target, nil
+	}
+	start, err := time.Parse(time.RFC3339, args.BaselineStartISO)
+	if err != nil {
+		return deviationAPIWindow{}, fmt.Errorf("baseline_start_time_iso must be RFC3339")
```
faiz9068: **Important — baseline bounds are parsed more strictly than the target window, and the error is unactionable.**

The target window resolves through `utils.GetTimeRangeAt` -> `utils.ParseToolTimestamp`, which accepts `2006-01-02 15:04:05` alongside RFC3339. The baseline uses a raw `time.Parse(time.RFC3339, ...)`. So `start_time_iso: "2026-07-15 11:45:00"` is accepted while an identically formatted `baseline_start_time_iso` is rejected — the same format valid in one argument and invalid in its sibling.

The message also drops both the offending value and the parse error, so a caller cannot tell which format was wanted. And the equal-duration/overlap arithmetic below then compares instants produced by two different parsers.

Fix: use `utils.ParseToolTimestamp` for both baseline bounds, and include the received value in the error.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,493 @@
...
+		if s.ServiceName != "" {
+			services[s.ServiceName] = true
+		}
+		if s.StatusCode == traceWaterfallErrorStatusCode {
+			resp.Data.Summary.ErrorCount++
+		}
+		if depth[id] > resp.Data.Summary.MaxDepth {
+			resp.Data.Summary.MaxDepth = depth[id]
+		}
+		if id == in.selectedID {
```
faiz9068, line 403: **Important — a `selected_span_id` that matches no returned span is a silent no-op.**

When no span matches, `SelectedSpan` stays nil and is omitted from the response. That makes three different situations indistinguishable to the model: a typo'd span ID, a span dropped by truncation / the environment filter / an unparseable timestamp, and a span that genuinely has no attributes or events. The likely next move is a retry with the same wrong ID.

`get_trace_waterfall.md` also presents its warning list as exhaustive, which is a genuinely nice property — worth preserving by emitting something like `selected_span_id not found in returned spans: <id>` and adding it to that list.

internal/prompts/descriptions/get_trace_attribute_deviations.md
```diff
@@ -0,0 +1,27 @@
...
+- `limit`: default 10, maximum 10.
+
+Omitted numeric bounds take the defaults above. A value outside a documented range is rejected with a message naming the bound, so a rejected call means the argument must be corrected — not retried unchanged.
+
+Reading `evidence.warnings`: `candidate_limit_reached` on a discovery call means discovery filled its server-side candidate budget, not that the request was wrong — pass explicit `candidate_attributes` to examine dimensions it did not reach. If the call fails with the capability being unavailable or disabled, this analysis is not enabled for the connected tenant; do not retry, and fall back to `get_traces` aggregations.
+
+The tool delegates to a bounded atomic trace-analysis endpoint: candidate discovery, cohort totals, value distributions, ranking, and representative trace IDs are resolved in a single server-side query with server-enforced limits. It returns full-denominator shares, percentage-point deltas, finite ratios, missing-value counts, exclusions, deterministic ranks, truncation/partial metadata, and evidence quality under `investigation-evidence/v1`.
```
faiz9068: **Important — this describes endpoint behavior the repo cannot verify and no test pins.**

Three claims are load-bearing for the model and unenforceable here: the `candidate_limit_reached` warning string (line 23), sensitive/identifier-like attributes failing the call rather than being dropped (line 12), and the field list on this line (full-denominator shares, percentage-point deltas, finite ratios, missing-value counts, exclusions, deterministic ranks).

Descriptions are behavior, not prose — the model acts on them. If the endpoint lands with different names, this becomes an actively misleading manual and nothing catches the drift.

Either tie these strings to the endpoint's own contract test via a shared fixture, or soften to what this client can guarantee and let `evidence.warnings` speak for itself.

contracts/feature-flags-v1.json
```diff
@@ -0,0 +1,12 @@
+{
+  "manifest_version": "investigation-feature-flags/v1",
+  "flags": [
+    { "name": "investigation.backend.attribute_deviations_v2", "default": false, "owner": "trace-api" },
+    { "name": "investigation.backend.direct_child_dependencies_v1", "default": false, "owner": "trace-api" },
+    { "name": "investigation.mcp.atomic_deviations", "default": false, "owner": "mcp" },
```
faiz9068: **Important — the flag manifest is inert, so the tool is served on the default surface while its capability is off.**

Nothing in the server reads this file; it is exercised only by a test. So the reserved default-off flag does not gate anything, and every client pays the description's token cost for a tool whose only outcome today is an error.

The 404 path does produce a "capability may be unavailable or disabled" hint, so the failure is survivable — this may well be intentional. Worth stating so explicitly, or gating registration behind config until the endpoint lands.

Separately: five of the seven flags have no consumer in this repo, and two are owned by another surface entirely (lines 8-9). Committing another surface's flag names here creates cross-surface coupling in this repo's artifacts, which cuts against the convention of not cross-referencing other surfaces from this codebase.

contracts/fixtures/evidence-trace-waterfall.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "evidence": {
+    "partial": false,
+    "truncated": false,
+    "warnings": [],
+    "returned_spans": 3,
+    "provenance": { "source": "trace-details", "observed_at": "2026-07-15T08:41:05Z", "query_id": "seed-waterfall-001" }
+  },
+  "interpretation": {
+    "evidence_quality": "medium",
```
faiz9068: **Minor — this fixture is not producer-shaped, so it cannot catch producer drift.**

`evidence_quality: "medium"` alongside `warnings: []` and `truncated: false` is a combination `buildTraceWaterfall` would never emit — that input yields `high`. The span entries also use `parent_span_id: null` where the producer omits the key (line 26), and the payload lacks `applied_limit`, `summary.start`/`end` and `summary.root_span_ids`.

Defensible as a cross-surface compatibility seed. But since `waterfall_contract_test.go` already marshals real producer output, generating this fixture from the producer would keep the two from diverging.

Related: `TestWorkflowEvidenceHashesMatchImmutableFixtures` compares fixture hashes to constants committed in the same change — a useful immutability tripwire, but worth labelling as such so it is not mistaken for conformance coverage.

internal/telemetry/traces/deviations.go
```diff
@@ -0,0 +1,268 @@
...
+type deviationAPICandidates struct {
+	Attributes   []string `json:"attributes"`
+	AutoDiscover bool     `json:"auto_discover"`
+}
+
+type deviationAPILimits struct {
+	MinimumCohortSize         int `json:"minimum_cohort_size"`
+	MinimumValueSupport       int `json:"minimum_value_support"`
+	MaximumCandidates         int `json:"maximum_candidates,omitempty"`
+	MaximumValuesPerAttribute int `json:"maximum_values_per_attribute,omitempty"`
```
faiz9068: **Minor — dead fields.**

`MaximumValuesPerAttribute` and `RepresentativesPerResult` are never assigned anywhere; they exist only so a test can assert they are absent from the wire. Since the endpoint owns those budgets — the point of `c544f65` — deleting the fields is cleaner than keeping fields whose only purpose is to be provably unset.

internal/telemetry/traces/deviations.go
```diff
@@ -0,0 +1,268 @@
...
+	limit := args.Limit
+	if limit == 0 {
+		limit = deviationRankedResultsDefault
+	} else if limit < 1 || limit > deviationRankedResultsMax {
+		return deviationAPILimits{}, fmt.Errorf("limit must be between 1 and %d", deviationRankedResultsMax)
+	}
+	return deviationAPILimits{
+		MinimumCohortSize:    cohort,
+		MinimumValueSupport:  support,
+		MaximumCandidates:    len(args.CandidateAttributes),
```
faiz9068: **Minor — tautological, and the same mirroring `c544f65` removed.**

`MaximumCandidates: len(args.CandidateAttributes)` tells the endpoint "at most exactly as many as I sent", which it can count from the list itself. Harmless on the wire, but it is still a client asserting a server-side scan budget.

internal/telemetry/traces/deviations.go
```diff
@@ -0,0 +1,268 @@
...
+
+func deviationTargetWindow(args GetTraceAttributeDeviationsArgs, now time.Time) (deviationAPIWindow, error) {
+	params := map[string]interface{}{}
+	if args.StartTimeISO != "" {
+		params["start_time_iso"] = args.StartTimeISO
+	}
+	if args.EndTimeISO != "" {
+		params["end_time_iso"] = args.EndTimeISO
+	}
+	if args.LookbackMinutes > 0 {
```
faiz9068: **Minor — negative `lookback_minutes` is silently ignored here.**

The `> 0` guard drops a negative value and falls through to the default window, whereas the waterfall path rejects it via `GetTimeRange`. Same argument, two behaviors.

Also: `lookback_minutes: 60` fails with "effective window must not exceed 15 minutes" without naming `lookback_minutes`, which slightly undercuts the description's promise that an out-of-range value is rejected with a message naming the bound.

internal/telemetry/traces/waterfall.go
```diff
@@ -0,0 +1,493 @@
...
+	SpanID             string                   `json:"span_id"`
+	ResourceAttributes map[string]string        `json:"resource_attributes,omitempty"`
+	SpanAttributes     map[string]interface{}   `json:"span_attributes,omitempty"`
+	Events             []map[string]interface{} `json:"events,omitempty"`
+	Links              []map[string]interface{} `json:"links,omitempty"`
+}
+
+// EvidenceWindow is a half-open [start,end) window.
+type EvidenceWindow struct {
+	Start    string `json:"start"`
```
faiz9068, line 73: **Minor — `start`/`end` lack `omitempty`.**

On an empty result, or one where every span was dropped, this emits `"start": ""` and `"end": ""`. Any consumer parsing them as date-time breaks on the empty string rather than on an absent key. Use `omitempty` or pointers.

Same shape: `EvidenceProvenance.QueryID` is always `null` from this producer.

Also nearby — `qp.Env` is set but never used in the request URL (environment filtering is client-side via `traceDetailsMatchesEnv`), which reads as if the server filters. Worth dropping or commenting.

And the served input schemas for both tools carry only `type` + `description` — no `minimum`/`maximum`/`format: date-time`. Bounds like `max_spans <= 1000` and `lookback_minutes <= 15` exist in prose only. That matches the struct-tag convention, but `get_apm_service_deviations` uses an explicit schema builder with bounds, which is the better precedent when the bounds are load-bearing.

faiz9068 (commented):
### Code review — trace investigation tools

Reviewed `740df9f3..a8558515` (merge-base with `main`). Verified on a clean checkout of the head commit: `go build ./...`, `go vet ./...`, `go test ./...` all clean across 16 packages, and `dump-tools --toolsets=logs` confirms neither new tool leaks outside the `traces` domain.

**What holds up under inspection, not just by claim:**

- `registerIfAllowed` wrapping is real — toolset filtering verified empirically, not just read.
- `waterfall_contract_test.go` validates *marshalled producer output* against `investigation-evidence-v1.schema.json`, including the populated, empty, truncated, orphan-parent and unparseable-timestamp paths. That closes the "stamp v1 and stay green" hole.
- Determinism is tested, not asserted: sorted ids/roots/children/warnings plus a 50-run byte-stability test. Matters because the contracts README hashes evidence.
- `self_time_ms` is an interval union with child clipping, and the test proves union rather than naive sum. Clipping warns instead of silently absorbing a malformed child.
- `parseTraceTimestampNano` returns `(int64, bool)` with `haveBounds` tracked separately, so instant 0 can no longer double as an unset sentinel.
- 5xx bodies provably do not reach the model — status is checked before the body is read, and `errors.go` relays a sanitized excerpt only for 400/422.
- Non-causal discipline is enforced structurally: the `claim_type` enum excludes `cause`, and `insufficient` for an empty trace is the right call over "trace does not exist".

**No Critical issues.** Six Important findings and a set of minor ones are left as inline comments. The one that blocks merge is the description-token-budget gate (see the comment on `dump_test.go`) — it is a repo-standard change rather than a code fix, so it needs an explicit decision.

internal/telemetry/traces/deviations.go
faiz9068: Fixed in 250246a. Both baseline bounds now go through `utils.ParseToolTimestamp`, the same helper the target window uses, so the asymmetry is gone — and the error quotes the value it received plus the underlying parse error.

Verified against a running server: `baseline_start_time_iso: "2026-07-30 10:00:00"` alongside an identically formatted `start_time_iso` now passes validation and reaches the network, where before it failed locally with "must be RFC3339". An unparseable value returns `baseline_start_time_iso "30/07/2026" is not a supported timestamp: unsupported time format ...`.

Tests cover three layouts (space-separated, RFC3339, RFC3339Nano) so the two arguments cannot drift apart again, plus one asserting the received value appears in the message.

internal/telemetry/traces/waterfall.go
faiz9068: Fixed in 46f6d16. An unmatched `selected_span_id` now emits `selected_span_id not found in returned spans: <id>`, and the warning list in the description — which does advertise itself as exhaustive — gained the string plus an explanation that the span may have been removed by truncation, the environment filter, or an unparseable timestamp, so the caller should not retry the same ID.

Confirmed on a live trace: a bogus ID returns no `selected_span` and the new warning, with `evidence_quality` dropping to `medium`; a valid ID returns the span and no warning. Both directions are pinned by a test.

internal/telemetry/traces/deviations.go
faiz9068: Fixed in 250246a. `lookback_minutes` is now rejected by name for negative and over-maximum values rather than being dropped by the `> 0` guard, so it matches the waterfall path.

Live: `lookback_minutes: -5` returns `lookback_minutes must be between 1 and 15, got -5`, and `60` returns the same message naming the argument — which is what the description promised and was not delivering.

internal/telemetry/traces/deviations.go
faiz9068: Fixed in 9b78293, taking the better option you named rather than the minimum.

The response is now rejected unless `contract_version` and `analysis_version` match the promised pair. On top of that, a verbatim copy of the endpoint's example response is committed as a fixture and served from an `httptest` server, so the handler's forwarded output is validated against `investigation-evidence-v1.schema.json` — mirroring `waterfall_contract_test.go`, which is the asymmetry you pointed at.

Two things worth noting. The check does not false-positive on real traffic: a live 200 from the endpoint carries both expected versions and passes. And a live response also validates against the evidence schema, so the conformance claim is now tested against the endpoint rather than only against a hand-written fixture.

The rejection half is covered by unit tests (wrong major version, wrong analysis version, missing envelope) since the real endpoint cannot be made to return a malformed shape on demand.

internal/prompts/descriptions/get_trace_attribute_deviations.md
faiz9068: Fixed in 9b78293 by pinning rather than softening, because the endpoint publishes an example response that both surfaces can test against. It is committed as `contracts/fixtures/evidence-attribute-deviations-endpoint.json`, and two tests hold the description to it: one asserting every field the description promises exists in it, one pinning the exact warning strings and `excluded_reason` values. A rename upstream now fails a test instead of leaving a misleading manual.

Checking the claims against the endpoint turned up three that were wrong, now corrected:

- `ratio` is nullable whenever either share is zero, so "finite ratios" was misleading. The description now says to fall back to `percentage_point_delta` and never read a null ratio as "no difference".
- `evidence_quality` is `medium` at best and `insufficient` when the evidence is partial. It is never `high`, which the description implied.
- Three emitted warnings were undocumented: `minimum_cohort_size_not_met`, `attribute_value_limit_reached`, and the `distinct_value_limit_exceeded` / `sensitive_or_high_cardinality_policy` exclusion reasons — which are `data.candidates[].excluded_reason` values, not warnings, and the description now keeps the two separate.

On one of the three you flagged, I think the description was already right: a named sensitive attribute does fail the call rather than being dropped, so I left that as-is.

I also removed the specific discovery-budget count I had briefly added. It is server-side tuning that can change without notice, and pinning it would make the description wrong the day it does — so the text now says only that the budget is small and that explicit `candidate_attributes` is how you look wider.

internal/telemetry/traces/deviations.go
faiz9068: Fixed in 250246a — both fields deleted, along with `maximum_candidates`.

Confirmed safe by reading the endpoint: it applies its own defaults before validating the request, so omitting all three yields the endpoint's budgets rather than a validation failure. Deleting is therefore strictly better than keeping fields whose only purpose was to be provably unset.

Note this supersedes my earlier reply on the limits comment, which said the two were "now populated at the bounds the endpoint applies by default". A later commit in this branch reverted that to let the endpoint own its budgets, so that reply was stale by the time you read the code.

internal/telemetry/traces/deviations.go
faiz9068: Fixed in 250246a — removed. You are right that it is the client asserting a server-side scan budget, and the endpoint can count the list it was sent.

One thing worth recording since it came up while checking this. The truncation warning fires when the discovered candidate count equals the candidate budget, and omitting the field means the endpoint applies its own default — which is small. So the warning will keep appearing on most discovery calls, and it does mean dimensions were never examined, not just that a limit was touched. Rather than raise the budget from the client, which would widen the server-side scan and override a deliberate tuning decision, the description now says what the warning implies and points at explicit `candidate_attributes` as the way to look wider. Raising the default is a separate conversation with whoever owns that cost.

internal/telemetry/traces/waterfall.go
faiz9068: Partly fixed in 46f6d16, with two deliberate divergences.

Fixed: `qp.Env` is no longer set, with a comment recording that the trace-details URL carries no environment parameter and filtering is client-side — good catch, it did read as if the server filtered.

Fixed differently than suggested: the field that actually emits `""` is on the summary, not `EvidenceWindow`. `EvidenceWindow` is always populated from a resolved time range, and the schema requires `start`/`end` with `format: date-time`, so `omitempty` there would risk the contract for no gain. `summary.start`/`end` are the ones only set when bounds exist, and those now have `omitempty`. Verified against an empty result: the summary keys come back without `start` or `end` rather than with empty strings.

Not changed: `provenance.query_id` emitting `null`. The endpoint's own response does the same, so diverging here would make the two producers inconsistent for a consumer reading the shared envelope. Happy to change it if you would rather both surfaces drop it, but that is a cross-surface decision rather than a fix on this side.

On the input schemas carrying no `minimum`/`maximum`/`format`: agreed that the explicit builder is the better precedent when bounds are load-bearing, and the bounds are now at least enforced by name at the handler rather than living only in prose. Leaving the schema-builder migration out of this PR unless you want it here.

contracts/fixtures/evidence-trace-waterfall.json
faiz9068: Fixed in d3e0533 the way you suggested: the fixture is now generated from the producer.

`TestWaterfallFixtureMatchesProducerOutput` holds a canonical input, marshals `buildTraceWaterfall`, and fails if the committed file has drifted. Regenerate with `LAST9_UPDATE_FIXTURES=1`. Every combination you flagged is gone as a result — quality is `high` for this input rather than `medium`, `parent_span_id` is omitted on the root, and `applied_limit`, `summary.start`/`end` and `root_span_ids` are present.

Regenerating also showed the old `self_time_ms: 630` on the root was simply wrong: with children at [120,940] and [980,1060] the interval union is 900, so self time is 550. A hand-written fixture had no way to catch that.

Content hash refreshed in `workflow-cases-v1.json`, and I labelled `TestWorkflowEvidenceHashesMatchImmutableFixtures` as an immutability tripwire pointing at where producer conformance is actually asserted, per your related note.

contracts/feature-flags-v1.json
faiz9068: Fixed in d3e0533, though not by gating registration.

Checking the owning API: the capability is enabled for all tenants, and when it is off the endpoint is not routed at all — which is the 404 the tool already turns into the "capability may be unavailable or disabled" hint. So unconditional registration is correct here and there is nothing to gate; the failure mode you described as survivable is the intended one.

The real problem was the manifest itself. The flag it named for this capability does not exist under that name, and of the seven entries only one gate exists at all. It now records that single gate, and the test requires every entry to declare who enforces it and whether this repo consumes it — so the file cannot drift back into reserving names nothing implements.

The flags owned by other surfaces are gone, which also settles the cross-surface coupling point.

dump_test.go
faiz9068: Owner decision: the removal was deliberate, and the gate stays out. Your read of the commit trail is accurate, so this is a "why", not a correction.

The reasoning, now recorded in `CONTEXT.md` in d3e0533 rather than left unwritten: a chars/4 ceiling on the summed served surface fails whenever a *correct* description grows, and correctness outranks size for this artifact. A wrong unit or a stale warning string changes model behaviour; a few hundred extra characters do not. This PR is a case in point — reviewing it surfaced a nullable field described as finite, an evidence-quality range that was overstated, and three undocumented warning strings, and fixing those honestly made the descriptions longer, not shorter.

You are right that nothing replaces it as an aggregate tripwire, and the CONTEXT.md entry says so explicitly rather than implying the whale and toolset mechanisms cover it — they bound per-tool size only. The deliberate position is that a description-heavy change is reviewed on whether every claim is true and load-bearing, not against a number. Restoring the ceiling would have meant trimming accurate text to fit, which is the outcome worth avoiding.

Recording the decision was the part that was genuinely missing, since the same commit deleted the CONTEXT.md decision block along with the test.

faiz9068 (approved):

PR #189, opened 2026-07-18
feat(mcp): toolsets + progressive disclosure for description token budget

prathamesh-sonpatki:
## Summary
- Add named **toolsets** (`logs`, `traces`, `metrics`, `alerts`, `dashboards`, `investigate`, `all`) that hard-filter `tools/list` via `--toolsets` / `LAST9_TOOLSETS` (unset = all).
- Move whale manuals (`get_logs` / `get_traces` / `get_service_logs`) to MCP resources (`last9://reference/*`) while keeping critical DSL/HTTP rules on-tool.
- Stop injecting org attribute catalogs (`{{labels}}`) into served descriptions.
- Enforce description-token ceilings in tests (`all` ≤ ~12k, `investigate` ≤ ~10k via chars/4).

Addresses ENG-1510 (automation hosts burning tens of thousands of tokens on `tools/list`).

Companion: last9-mcp-evals PR for `*_base.md` eval paths.

## Test plan
- [x] `go test ./...` (and focused toolsets/dump/prompts tests)
- [x] `dump-tools` / `dump-tools --toolsets=investigate` budget smoke
- [x] last9-mcp-evals `--use-server`: **80/80** (log, trace, log_tool_selection, alerts, exceptions, datasource, type_coercion)
- [x] Thermo-nuclear review: no customer/competitor names or secrets in PR tree
- [ ] Reviewer: spot-check `resources/list` + `resources/read` for `last9://reference/logjson`


prathamesh-sonpatki:
Companion evals PR: https://github.com/last9/last9-mcp-evals/pull/31

Thermo-nuclear review (both PRs): hard gate clean — no customer refs, competitor refs, or secrets in the shipped diffs. Remote tip rewritten so Tenderd/Bifrost never appear on the force-pushed history.

kneurgao (commented):
10 new findings:
- P1 — `logs` toolset excludes `prometheus_instant_query`, which its own reference manuals require (`internal/toolsets/toolsets.go:14`)
- P1 — `get_traces_base.md` drops aggregation-order + exact key-name rules, now resource-only (`internal/prompts/descriptions/get_traces_base.md:19`)
- P1 — tool-registration error propagation is dead code; SDK panics on bad schema, uncaught (`tools.go:25`)
- P2 — four `go:embed` description vars now dead code (`internal/prompts/prompts.go:6`)
- P2 — `toolsets.Parse("all,<unknown>")` silently accepts the bad token (`internal/toolsets/toolsets.go:117`)
- P2 — `get_service_traces` excluded from whale→resource split, base file orphaned (`tools.go:44`)
- P2 — `get_logs`/`get_traces` lack a critical-rule content regression test (`internal/prompts/prompts_test.go:31`)
- P2 — attribute-cache warm/refresh subsystem now has zero consumers (`main.go:226`)
- P2 — `resources/read` not-found path is untested (`resources.go:48`)
- P2 — tracejson/service_logs resource bodies never read in tests (`resources_test.go:57`)

internal/toolsets/toolsets.go
```diff
@@ -0,0 +1,155 @@
...
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+)
+
+// Named toolset → tool membership. "all" and "investigate" are composites
+// handled in Parse, not listed here as flat membership.
+var named = map[string][]string{
+	"logs": {
```
kneurgao, line 14: **[P1]** `logs` toolset excludes `prometheus_instant_query`, which `logjson.md`/`service_logs.md` both instruct the model to call first ("discover log-sending services before a broad search") whenever the target service isn't named. Under `--toolsets logs`, an agent following its own on-tool/resource guidance hits a tool that isn't in `tools/list`, with no substitute inside this toolset. `prometheus_instant_query` only lives in `metrics`/`investigate`. Either add it (or a narrower log-index-discovery primitive) to `logs`, or trim the service-inventory step out of the logs reference docs for logs-only deployments.

internal/prompts/descriptions/get_traces_base.md
```diff
@@ -1,34 +1,21 @@
...
+**Time (tool args):** Prefer `lookback_minutes` (default **5**). Absolute → `start_time_iso`+`end_time_iso` (RFC3339). Never put the window as a `Timestamp` filter in the pipeline.
 
-Returns comprehensive trace data including trace IDs, spans, durations, timestamps, and metadata.
+**Fields:** Top-level: `TraceId`, `SpanId`, `ServiceName`, `SpanName`, `SpanKind`, `StatusCode`, `Duration`, `Timestamp`, `ParentSpanId`. SpanKind/StatusCode need full OTel prefixes (`SPAN_KIND_SERVER`, `STATUS_CODE_ERROR`—not `SERVER`/`ERROR`). **`Duration` is nanoseconds** (1000ms = `1000000000`). Span/resource attrs → `get_trace_attributes*`; use `attributes['key']` / `resources['key']` (never `SpanAttributes.foo`).
 
-IMPORTANT: There is no "filter_tags", "tags", or "attributes" parameter. ALL filtering — including by span tags,
-attributes, session IDs, trace metadata, or any key-value pair — must be expressed as a tracejson_query filter.
-Do NOT invent parameter names; use tracejson_query exclusively for filtering.
+**Order:** filter first (match-all on `TraceId`/`SpanId` if needed before aggregate). Show/find → filter only; aggregate/window_aggregate only for counts/analysis.
```
kneurgao: **[P1]** This on-tool description keeps field-shape/OTel-prefix rules but drops two rules from `tracejson.md` that most directly cause malformed-query 400s for a client that never calls `resources/read`: (1) aggregation must always be preceded by a filter stage, never start a pipeline with `aggregate`, and (2) the exact key names are `aggregates`/`groupby`, not `aggregations`/`group_by`. This is the same class of regression this branch already fixed twice (`633bf29` "restore critical pipeline rules in whale base descriptions", `9d75ef6` "tighten absolute-time, HTTP status, and l9_sanity rules") — worth folding a compressed version of both rules back on-tool; there's headroom under the 2000-char tripwire.

tools.go
```diff
@@ -14,286 +14,294 @@ import (
...
-// base tool description. For get_logs, it also replaces the {{labels}} placeholder
-// with the actual attribute names from the cache.
-func buildEnhancedDescription(base, instructions string, labelValues []string) string {
-	desc := base + "\n\n" + instructions
-	if len(labelValues) > 0 {
-		desc = strings.ReplaceAll(desc, "{{labels}}", strings.Join(labelValues, ", "))
-	} else {
-		desc = strings.ReplaceAll(desc, "{{labels}}", "")
+// registerIfAllowed registers a tool only when it is in the active toolset set.
+func registerIfAllowed[In, Out any](server *last9mcp.Last9MCPServer, allowed toolsets.Set, tool *mcp.Tool, handler mcp.ToolHandlerFor[In, Out]) error {
```
kneurgao: **[P1]** `registerIfAllowed` now checks the return value of `last9mcp.RegisterInstrumentedTool`, but that function (vendored `mcp-go-sdk@v0.1.2`, `middleware.go:400-403`) unconditionally returns `nil` — it never surfaces an error. The real failure mode lives one level deeper: `modelcontextprotocol/go-sdk@v1.5.0`'s `AddTool` **panics** on a schema error instead of returning one, and there's no `recover()` anywhere in `main.go`/`tools.go`/`dump.go`. A malformed `InputSchema` on any tool (or a future handler) will hard-crash the process with a raw panic, not the clean `log.Fatalf` this PR's "address thermo review" commit implies now exists. Wrap `registerAllTools()`/`dumpTools()` in a `recover()` that funnels into the existing error path, and/or fix upstream `mcp-go-sdk` to surface `AddTool`'s error instead of swallowing it.

internal/prompts/prompts.go
```diff
@@ -3,22 +3,34 @@ package prompts
 import _ "embed"
 
 //go:embed descriptions/get_logs.md
-var GetLogsInstructions string
+var GetLogsInstructions string // stub; long manual is LogjsonReference
```
kneurgao: **[P2]** `GetLogsInstructions`, `GetTracesInstructions`, `GetMetricsInstructions`, `GetServiceLogsInstructions` (lines 6, 9, 15, 21) have zero call sites anywhere in the repo post-PR — `tools.go` now builds these descriptions from the `*Description` vars instead. Delete the four vars, their `go:embed` directives, and the now-purposeless `get_logs.md`/`get_traces.md`/`get_metrics.md`/`get_service_logs.md` breadcrumb files (the migration note can live in a comment or `CONTEXT.md`).

internal/toolsets/toolsets.go
```diff
@@ -0,0 +1,155 @@
...
+		if t == "" {
+			continue
+		}
+		tokens = append(tokens, strings.ToLower(t))
+	}
+	if len(tokens) == 0 {
+		return nil, nil
+	}
+
+	for _, t := range tokens {
```
kneurgao, line 130: **[P2]** `Parse` checks each token for the literal `"all"` and returns `(nil, nil)` the instant it finds one — before the validation loop that checks unknown names ever runs. So `--toolsets=all,typo` silently returns the full unfiltered surface instead of erroring, exactly the "silent fallback to `all`" behavior `CONTEXT.md` (added by this same PR) explicitly lists as something to avoid. `TestParseAllSupersedes` only covers `all,logs` (both valid). Validate every token against `named`/`investigate` first (or independently of the `all` short-circuit), then apply the short-circuit only once every token is confirmed valid.

tools.go
```diff
@@ -14,286 +14,294 @@ import (
...
-	getLogsDesc := buildEnhancedDescription(prompts.GetLogsDescription, prompts.GetLogsInstructions, attrCache.GetLogAttributes())
-	getServiceLogsDesc := buildEnhancedDescription(prompts.GetServiceLogsDescription, prompts.GetServiceLogsInstructions, attrCache.GetLogAttributes())
-	getTracesDesc := buildEnhancedDescription(prompts.GetTracesDescription, prompts.GetTracesInstructions, nil)
-	getServiceTracesDesc := buildEnhancedDescription(prompts.GetServiceTracesDescription, prompts.GetServiceTracesInstructions, nil)
-	getMetricsDesc := buildEnhancedDescription(prompts.PromqlRangeQueryDetails, prompts.GetMetricsInstructions, nil)
+	// Whales: short on-tool description only (manuals are MCP resources).
+	getLogsDesc := prompts.GetLogsDescription
+	getServiceLogsDesc := prompts.GetServiceLogsDescription
+	getTracesDesc := prompts.GetTracesDescription
+	getServiceTracesDesc := prompts.GetServiceTracesInstructions
```
kneurgao: **[P2]** The comment above ("Whales: short on-tool description only, manuals are MCP resources") covers this block, but `getServiceTracesDesc` is set from `prompts.GetServiceTracesInstructions` (full text), not a shrunk `_base.md` stub — unlike its four siblings. `prompts.GetServiceTracesDescription` (the base var, `prompts.go:153`) is now unreferenced by anything, and no `last9://reference/service_traces` resource exists in `resources.go`, so a future editor updating the base file would silently have zero effect on `tools/list` — the exact "parallel `.md` copy drifts" failure `AGENTS.md`'s markdown-single-source rule was written to prevent. Either migrate `get_service_traces` to the same base+resource split, or delete the orphaned base file/var and fix the comment's scope.

internal/prompts/prompts_test.go
```diff
@@ -7,26 +7,51 @@ import (
...
 	for _, c := range checks {
-		if !strings.Contains(inst, c.phrase) {
-			t.Errorf("GetServiceLogsInstructions missing %q: %s", c.phrase, c.reason)
+		if !strings.Contains(desc, c.phrase) {
+			t.Errorf("GetServiceLogsDescription missing %q: %s", c.phrase, c.reason)
+		}
+	}
+}
+
+func TestWhaleDescriptionsBounded(t *testing.T) {
```
kneurgao, line 77: **[P2]** `TestGetServiceLogsDescriptionCriticalRules` pins specific must-contain phrases for `get_service_logs`, but `TestWhaleDescriptionsBounded` only checks `get_logs_base`/`get_traces_base` for non-empty + ≤2000 chars — no content assertions. Both files carry narrow, easily-droppable critical rules (e.g. the `$exists` ban, the HTTP-5xx-status-code rule, the exact-`trace_id`-omit-`service_name` rule) that a future edit could silently strip without failing CI, since dropping one sentence won't trip a length ceiling. Mirror the `get_service_logs` pattern for `get_logs`/`get_traces`.

main.go
```diff
@@ -185,12 +216,14 @@ func main() {
...
 
+	registerReferenceResources(server)
+
 	// Register all tools
 	if err := registerAllTools(server, cfg, attrCache); err != nil {
 		log.Fatalf("failed to register tools: %v", err)
 	}
 
-	// Background goroutine to refresh attributes and re-register tools periodically
+	// Background goroutine keeps the attribute cache warm for handlers that may use it.
```
kneurgao: **[P2]** `buildEnhancedDescription` (the only caller of `attrCache.GetLogAttributes()`) is deleted by this PR, but the startup warm-fetch and this 2h refresh goroutine are still wired through `main.go`/`dump.go`/`tools.go` (`tools.go:37` marks the param dead with `_ = attrCache`). The goroutine no longer even re-registers tools after refreshing (per commit `2cd1f76`), so it now does nothing but log a line every 2 hours. Delete `AttributeCache` end-to-end, or land its real consumer in this PR.

resources.go
```diff
@@ -0,0 +1,58 @@
...
+					Contents: []*mcp.ResourceContents{{
+						URI:      ref.uri,
+						MIMEType: "text/markdown",
+						Text:     ref.body,
+					}},
+				}, nil
+			}
+		}
+		return nil, mcp.ResourceNotFoundError(req.Params.URI)
+	}
```
kneurgao, line 48: **[P2]** The not-found fallback (`mcp.ResourceNotFoundError`) is never exercised — `resources_test.go` only calls `ReadResource` with known URIs. Add a subtest calling `ReadResource` with an unknown URI (e.g. `last9://reference/does-not-exist`) and asserting an error.

resources_test.go
```diff
@@ -0,0 +1,78 @@
...
+	defer session.Close()
+
+	found := map[string]bool{}
+	for res, err := range session.Resources(ctx, nil) {
+		if err != nil {
+			t.Fatal(err)
+		}
+		found[res.URI] = true
+	}
+	for _, uri := range []string{resourceURILogjson, resourceURITracejson, resourceURIServiceLogs, resourceURIMetrics} {
```
kneurgao, line 55: **[P2]** `TestReferenceResourcesAvailableUnderMetricsToolset` only checks `resources/list` presence for `tracejson`/`service_logs`, and only does a real `ReadResource` + body-length round-trip for `logjson`/`metrics`. The PR's own test plan flagged the logjson spot-check as an open TODO — that's closed, but the same read-path check was never extended to the other two moved manuals, so a mis-wired URI-to-body mapping there wouldn't be caught. Extend the read-path assertions to all four URIs.

kneurgao (commented):
No new issues. All 10 previously flagged items look addressed and tested — mind marking those threads resolved?

kneurgao (approved):
Looks good — approved.

PR #192, opened 2026-07-21
Build multi-arch (amd64+arm64) mcp-server image

kneurgao:
## Summary

- **Dockerfile**: builder is now `FROM --platform=$BUILDPLATFORM golang:1.25-alpine` with `ARG TARGETOS`/`ARG TARGETARCH`; the pure-Go (`CGO_ENABLED=0`) binary is built with `GOOS/GOARCH`, so it cross-compiles on the native runner and the Go compile is never emulated. buildx pulls the final alpine stage for the target platform (its trivial `apk`/`adduser`/`chown` RUN steps run under QEMU).
- **.github/workflows/release.yaml** (`publish-docker`): replace plain `docker build`/`tag`/`push` with `docker/setup-qemu-action@v3` + `docker/setup-buildx-action@v3` + `docker/build-push-action@v6`, pushing `linux/amd64,linux/arm64` as a manifest list with `provenance: false`. Tag scheme unchanged (`:latest`, `:v<ver>`, `:<branch>`, `:<branch>-v<ver>`) and same `VERSION`/`COMMIT`/`BUILD_TIME` build-args.

## Why

`mcp-server` currently ships an amd64-only image, forcing consumers onto amd64 nodes. On the `credit-euc1-tdp` (cpluseuc1) Karpenter/Graviton cluster — whose default nodepools prefer cheaper arm64 instances — the amd64-only image lands on an arm64 node and dies with `exec format error`. This is worked around today by a `node_selector = { "kubernetes.io/arch" = "amd64" }` pin in the shipment repo. Making the image multi-arch lets it run natively on Graviton and lets that pin be dropped.

Mirrors the established multi-arch pattern in `last9/last9` (#11006 otel-ingestor, #11001 scheduled-jobs) and PDE-381 (configmapupdater).

## Test plan

- [x] `make`-equivalent cross-compile locally: `GOOS=linux GOARCH=amd64` and `GOARCH=arm64` both produce static ELF binaries (x86-64 / aarch64).
- [x] `docker buildx build --platform linux/amd64,linux/arm64 --provenance=false` produces a **2-manifest list** (`linux/amd64`, `linux/arm64`) — verified via OCI export.
- [ ] CI on a `release-*` branch pushes a multi-arch tag; verify with `docker buildx imagetools inspect 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:<TAG>` → two manifests.
- [ ] Existing amd64 consumers pull/run the new tag unchanged.
- [ ] After a multi-arch tag reaches cpluseuc1, remove the `node_selector` amd64 block in shipment and confirm the pod schedules + runs on an arm64 node.

Companion `supervisor-mithai` is blocked on its `mithai` base image (itself amd64-only) — tracked in ENG-1529.

Closes ENG-1528

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -179,6 +179,12 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
 
+      # setup-buildx-action creates a docker-container buildkit driver (separate
+      # from the host daemon). The multi-arch push below works because the ECR
+      # login step above writes a plain base64 `auth` into ~/.docker/config.json,
+      # which build-push-action forwards into the builder. If amazon-ecr-login is
+      # ever switched to a credential-helper (credsStore) mode, that auth won't
+      # propagate to the builder and the push will start failing with 401s.
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
@@ -199,6 +205,8 @@ jobs:
             VERSION=${{ steps.get_version.outputs.version }}
             COMMIT=${{ github.sha }}
             BUILD_TIME=${{ steps.build_time.outputs.build_time }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
           tags: |
             652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:${{ steps.extract_branch.outputs.branch_name }}
             652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:${{ steps.extract_branch.outputs.branch_name }}-v${{ steps.get_version.outputs.version }}
@@ -216,6 +224,8 @@ jobs:
             VERSION=${{ needs.tag-and-release.outputs.new_version }}
             COMMIT=${{ github.sha }}
             BUILD_TIME=${{ steps.build_time.outputs.build_time }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
           tags: |
             652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:v${{ needs.tag-and-release.outputs.new_version }}
             652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:latest
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -173,39 +173,49 @@ jobs:
       - name: Login to Amazon ECR
         uses: aws-actions/amazon-ecr-login@v2
 
+      # QEMU registers binfmt handlers so the arm64 leg's alpine RUN steps
+      # (apk/adduser/chown) can execute; the Go compile itself cross-compiles
+      # on the native amd64 runner and is never emulated.
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Push to Amazon ECR (Release Branch)
         if: |
           github.event_name == 'push' &&
           startsWith(github.ref, 'refs/heads/release-')
-        env:
-          VERSION: ${{ steps.get_version.outputs.version }}
-          COMMIT: ${{ github.sha }}
-          BRANCH: ${{ steps.extract_branch.outputs.branch_name }}
-          BUILD_TIME: ${{ steps.build_time.outputs.build_time }}
-        run: |
-          docker build -f Dockerfile \
-            --build-arg VERSION="${VERSION}" \
-            --build-arg COMMIT="${COMMIT}" \
-            --build-arg BUILD_TIME="${BUILD_TIME}" \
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -173,39 +173,49 @@ jobs:
...
-          docker build -f Dockerfile \
-            --build-arg VERSION="${VERSION}" \
-            --build-arg COMMIT="${COMMIT}" \
-            --build-arg BUILD_TIME="${BUILD_TIME}" \
-            -t 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:build .
-          docker tag 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:build 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:${BRANCH}
-          docker tag 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:build 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:${BRANCH}-v${VERSION}
-          docker push 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:${BRANCH}
-          docker push 652845092827.dkr.ecr.ap-south-1.amazonaws.com/mcp-server:${BRANCH}-v${VERSION}
+        uses: docker/build-push-action@v6
```
kneurgao: On every release this compiles once per arch with a cold Docker layer cache — the `go mod download` deps layer isn't reused run-to-run (and `go build -a` defeats Go's own cache). Not a regression (the old `docker build` had no cache either), but a GHA layer cache reclaims it. Will add `cache-from: type=gha` / `cache-to: type=gha,mode=max` here and on the Main-Branch step.

.github/workflows/release.yaml
```diff
@@ -173,39 +173,49 @@ jobs:
...
         uses: aws-actions/amazon-ecr-login@v2
 
+      # QEMU registers binfmt handlers so the arm64 leg's alpine RUN steps
+      # (apk/adduser/chown) can execute; the Go compile itself cross-compiles
+      # on the native amd64 runner and is never emulated.
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
kneurgao: setup-buildx-action creates a `docker-container` buildkit driver, separate from the host daemon. The multi-arch push to ECR works only because `amazon-ecr-login@v2` writes a plain base64 `auth` into `~/.docker/config.json` (not a `credsStore` helper), which build-push-action forwards into the builder. If ECR login is ever switched to credential-helper mode this push will 401. Will add a one-line comment here capturing that assumption.

.github/workflows/release.yaml
kneurgao: Fixed in 83b0ec7 — added `cache-from: type=gha` / `cache-to: type=gha,mode=max` to both the Release-Branch and Main-Branch build-push steps.

.github/workflows/release.yaml
kneurgao: Fixed in 83b0ec7 — added a comment above the *Set up Docker Buildx* step documenting the docker-container driver's reliance on the plain `auth` that amazon-ecr-login writes to ~/.docker/config.json (credsStore mode would 401).

kneurgao:
### Reworked to drop QEMU (per ENG-1074) — pushed in f7f4f6d

Original revision used `docker/setup-qemu-action` + `build-push-action --platform linux/amd64,linux/arm64`, i.e. an **emulated** arm64 build. ENG-1074 flags that qemu cross-arch docker builds reliably time out on hosted runners, so this now follows the native-runner **D3 pattern** (ENG-1049) instead:

- **`build-docker`** — matrix job, one arch per **native** runner: `amd64` → `ubuntu-latest`, `arm64` → `ubuntu-24.04-arm` (the same hosted arm64 runner `mithai-internal` already uses). Each pushes a per-arch staging tag `:sha-<short>-<arch>`. Nothing is emulated.
- **`docker-manifest`** — merges the two staging images into the real tags with `docker buildx imagetools create` (a registry-side manifest-list write — no build, no qemu, no builder).

Tag scheme unchanged (`:latest`, `:v<ver>`, `:<branch>`, `:<branch>-v<ver>`). `setup-qemu-action`/`setup-buildx-action` and the GHA layer cache are gone (the container-driver cache no longer applies). Dockerfile keeps `FROM --platform=$BUILDPLATFORM` + `ARG TARGETARCH`, so `BUILDPLATFORM == TARGETPLATFORM` on the native build.

Verified locally: native `docker build --platform linux/arm64 --build-arg TARGETARCH=arm64` produces a `linux/arm64` image whose binary runs and reports the injected version.

_Note: per-arch staging tags (`:sha-<short>-*`) remain in ECR after the merge — same as the D3 reference; can be pruned via an ECR lifecycle rule if desired._

prathamesh-sonpatki (approved):

PR #195, opened 2026-07-22
fix(traces): never chunk aggregate pipelines; rewrite $exists to the working existence idiom

prathamesh-sonpatki:
Fixes FDE-285 and FDE-286.

## Problem

1. **FDE-285** — `get_traces` split long time ranges into chunks even when the pipeline contained an `aggregate`/`window_aggregate` stage, then concatenated per-chunk responses. The backend merges group-bys only within one request, so the same group key appeared once per chunk (duplicate rows with split counts), and non-decomposable aggregates (`$avg`, `$median`, `$quantile`, `$stddev`) were mathematically wrong, not just duplicated. A 7-day aggregate query ran as 192 backend requests; a single failed chunk yielded silently incomplete groups. The dashboard never hits this — it sends one un-chunked request.
2. **FDE-286** — the sanitizer allowlisted `$exists` (and recommended it in its error message), but the backend has no such operator: unknown operators compile to `1=1`, so the condition was silently dropped and matched every span.

## Changes

- `internal/utils/pipeline_inspect.go`: shared `PipelineHasAggregateStage` helper (`aggregate` / `window_aggregate`).
- `internal/telemetry/traces/traces.go`: aggregate pipelines short-circuit to a single full-window request (mirrors the existing exact-trace-id path); 408 responses now tell the caller to narrow the time window.
- `internal/telemetry/logs/logs.go`: delegates its existing no-chunk-for-aggregates check to the shared helper (no behavior change — logs never had the bug).
- `internal/telemetry/traces/query_sanitize.go`: `$exists` is rewritten in place to the working idiom `{"$neq": [field, ""]}` (recurses `$and`/`$or`/`$not`); removed from the allowlist; error message now teaches the idiom.
- `internal/telemetry/traces/input_schema.go`: existence hint no longer advertises the backend-broken `$notnull`; points to `$neq`-empty-string (backend fix for `$notnull` is in flight separately as ENG-1537).
- `internal/prompts/descriptions/get_traces.md`: documents the `_last9_mcp.partial_result` annotation.

## Verification

- `go test -race ./...` green; new tests assert exactly one upstream request for aggregate pipelines over a 7-day window (and >1 for plain filters), plus `$exists` rewrite coverage (top-level, nested, arg-less rejection).
- Live check against production data: a 7-day group-by-session aggregate that previously returned duplicate keys and `partial_result` now issues a single request (chunking debug log confirms) and returns 54 rows / 54 distinct keys / 0 dupes.
- `eval:trace` suite from last9-mcp-evals against this checkout: 23/23 pass.

### DIFF
--- a/internal/telemetry/traces/input_schema.go
+++ b/internal/telemetry/traces/input_schema.go
@@ -66,7 +66,7 @@ func filterStageSchema() map[string]interface{} {
 					"Numeric: $gt, $lt, $gte, $lte. " +
 					"String: $contains, $notcontains, $icontains, $inotcontains, $containsWords, $notcontainsWords, $icontainsWords, $inotcontainsWords. " +
 					"Regex: $regex, $notregex, $iregex, $inotregex. " +
-					"Existence: $notnull. " +
+					"Existence: no $notnull operator — use $neq against an empty string instead, e.g. {\"$neq\": [field, \"\"]}. " +
 					"Each operator takes [field, value] except logical operators which take an array of conditions. " +
 					"Example: {\"$and\": [{\"$eq\": [\"ServiceName\", \"api\"]}, {\"$eq\": [\"StatusCode\", \"STATUS_CODE_ERROR\"]}]}",
 			},
--- a/internal/telemetry/traces/query_sanitize.go
+++ b/internal/telemetry/traces/query_sanitize.go
@@ -10,7 +10,6 @@ var traceFilterFieldOperators = map[string]struct{}{
 	"$contains":          {},
 	"$containsWords":     {},
 	"$eq":                {},
-	"$exists":            {},
 	"$gt":                {},
 	"$gte":               {},
 	"$icontains":         {},
@@ -51,6 +50,12 @@ func sanitizeTraceJSONQuery(stages []map[string]interface{}) error {
 
 		if stageType == "filter" || stageType == "where" {
 			if query, ok := stage["query"]; ok {
+				// The backend has no $exists operator — an unknown operator
+				// compiles to `1=1` and silently matches every span instead of
+				// erroring, so rewrite it to the working idiom before
+				// validation ever sees it. Covers $exists nested inside
+				// $and/$or/$not since it walks the same condition tree.
+				rewriteExistsOperator(query)
 				if err := validateTraceFilterCondition(query, path+".query"); err != nil {
 					return err
 				}
@@ -107,6 +112,37 @@ func wrapTopLevelFilterQuery(query interface{}) interface{} {
 	return map[string]interface{}{"$and": conditions}
 }
 
+// rewriteExistsOperator walks a filter condition tree and rewrites every
+// $exists leaf ({"$exists": [field]}) in place to the working existence idiom
+// {"$neq": [field, ""]}. The backend has no $exists operator; passing it
+// through compiles to `1=1` and silently matches every span rather than
+// erroring, so models' natural "$exists" output is normalized here instead of
+// rejected. Recurses through $and/$or/$not so nested conditions are covered.
+func rewriteExistsOperator(value interface{}) {
+	switch typed := value.(type) {
+	case []interface{}:
+		for _, item := range typed {
+			rewriteExistsOperator(item)
+		}
+	case map[string]interface{}:
+		for key, item := range typed {
+			if _, isLogical := traceFilterLogicalOperators[key]; isLogical {
+				rewriteExistsOperator(item)
+				continue
+			}
+			if key != "$exists" {
+				continue
+			}
+			args, ok := item.([]interface{})
+			if !ok || len(args) == 0 {
+				continue
+			}
+			delete(typed, "$exists")
+			typed["$neq"] = []interface{}{args[0], ""}
+		}
+	}
+}
+
 func validateTraceFilterCondition(value interface{}, path string) error {
 	switch typed := value.(type) {
 	case []interface{}:
@@ -127,7 +163,7 @@ func validateTraceFilterCondition(value interface{}, path string) error {
 				continue
 			}
 			return fmt.Errorf(
-				"invalid filter condition key %q at %s: keys must be operators ($eq, $neq, $gt, $gte, $lt, $lte, $contains, $notcontains, $icontains, $inotcontains, $icontainsWords, $inotcontainsWords, $regex, $notregex, $iregex, $inotregex, $ieq, $ineq, $notnull, $exists, $containsWords, $notcontainsWords) or logical operators ($and, $or, $not); use the form {%q: [field, value]} — for example {\"$eq\": [\"ServiceName\", \"checkout\"]}",
+				"invalid filter condition key %q at %s: keys must be operators ($eq, $neq, $gt, $gte, $lt, $lte, $contains, $notcontains, $icontains, $inotcontains, $icontainsWords, $inotcontainsWords, $regex, $notregex, $iregex, $inotregex, $ieq, $ineq, $notnull, $containsWords, $notcontainsWords) or logical operators ($and, $or, $not); use the form {%q: [field, value]} — for example {\"$eq\": [\"ServiceName\", \"checkout\"]}. For existence checks use {\"$neq\": [field, \"\"]} — there is no $exists operator.",
 				key,
 				path,
 				"$eq",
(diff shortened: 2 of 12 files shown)


internal/telemetry/traces/input_schema.go
```diff
@@ -66,7 +66,7 @@ func filterStageSchema() map[string]interface{} {
 					"Numeric: $gt, $lt, $gte, $lte. " +
 					"String: $contains, $notcontains, $icontains, $inotcontains, $containsWords, $notcontainsWords, $icontainsWords, $inotcontainsWords. " +
 					"Regex: $regex, $notregex, $iregex, $inotregex. " +
-					"Existence: $notnull. " +
+					"Existence: no $notnull operator — use $neq against an empty string instead, e.g. {\"$neq\": [field, \"\"]}. " +
```
faiz9068, line 69: **Important — `$notnull` left silently broken at the server layer.** This hint correctly stops advertising `$notnull`, but `$notnull` is still in the `traceFilterFieldOperators` allow-list (`query_sanitize.go:29`) and is neither rewritten nor rejected. Since `$notnull` *was* the recommended existence idiom, models will keep emitting it — and it passes validation and reaches the backend, where it matches nothing (ENG-1537, verified in querygen.go). That's the same silent-match-nothing class this PR set out to kill. Suggest handling it like `$exists`: rewrite `$notnull` → `{"$neq": [field, ""]}` in `rewriteExistsOperator` (and rename accordingly), or drop it from the allow-list so it's rejected with the `$neq` guidance. A regression test mirroring the `$exists` tests would lock it in.

internal/telemetry/traces/query_sanitize.go
```diff
@@ -127,7 +163,7 @@ func validateTraceFilterCondition(value interface{}, path string) error {
 				continue
 			}
 			return fmt.Errorf(
-				"invalid filter condition key %q at %s: keys must be operators ($eq, $neq, $gt, $gte, $lt, $lte, $contains, $notcontains, $icontains, $inotcontains, $icontainsWords, $inotcontainsWords, $regex, $notregex, $iregex, $inotregex, $ieq, $ineq, $notnull, $exists, $containsWords, $notcontainsWords) or logical operators ($and, $or, $not); use the form {%q: [field, value]} — for example {\"$eq\": [\"ServiceName\", \"checkout\"]}",
+				"invalid filter condition key %q at %s: keys must be operators ($eq, $neq, $gt, $gte, $lt, $lte, $contains, $notcontains, $icontains, $inotcontains, $icontainsWords, $inotcontainsWords, $regex, $notregex, $iregex, $inotregex, $ieq, $ineq, $notnull, $containsWords, $notcontainsWords) or logical operators ($and, $or, $not); use the form {%q: [field, value]} — for example {\"$eq\": [\"ServiceName\", \"checkout\"]}. For existence checks use {\"$neq\": [field, \"\"]} — there is no $exists operator.",
```
faiz9068: **Important — contradictory guidance shipped.** This error message still lists `$notnull` among the valid operators, while the input schema (`input_schema.go:69`) now tells the model there is no `$notnull` operator. A model that trips an unrelated validation error is told `$notnull` is fine, directly contradicting the schema. Remove `$notnull` from this operator list and resolve it together with the allow-list handling above.

internal/telemetry/traces/query_sanitize.go
```diff
@@ -107,6 +112,37 @@ func wrapTopLevelFilterQuery(query interface{}) interface{} {
 	return map[string]interface{}{"$and": conditions}
 }
 
+// rewriteExistsOperator walks a filter condition tree and rewrites every
+// $exists leaf ({"$exists": [field]}) in place to the working existence idiom
+// {"$neq": [field, ""]}. The backend has no $exists operator; passing it
+// through compiles to `1=1` and silently matches every span rather than
+// erroring, so models' natural "$exists" output is normalized here instead of
+// rejected. Recurses through $and/$or/$not so nested conditions are covered.
+func rewriteExistsOperator(value interface{}) {
```
faiz9068: **Minor — unguarded sibling `$neq` clobber.** For a single condition object carrying both keys, e.g. `{"$exists": [f1], "$neq": [f2, ""]}`, this does `typed["$neq"] = []interface{}{args[0], ""}` and silently overwrites the original `$neq`, dropping the `f2` condition. Very unlikely from real model output (one operator per object), hence Minor — but a defensive check (skip if `$neq` already present, or fold both into an `$and`) removes the sharp edge.

internal/telemetry/traces/traces.go
```diff
@@ -287,6 +303,10 @@ func executeTraceJSONQuery(ctx context.Context, client *http.Client, cfg models.
 		if len(bodyStr) > 100 {
 			bodyStr = bodyStr[:100] + "... (truncated)"
 		}
+		if resp.StatusCode == http.StatusRequestTimeout {
```
faiz9068, line 306: **Minor — 408 wording is aggregate-specific but the branch isn't.** This lives in the shared `executeTraceJSONQuery`, which also serves the exact-trace-id lookup and every *per-chunk* call of a non-aggregate query. A 408 on an already-narrow chunk surfaces "trace **aggregation** query timed out … narrow the time window" — inaccurate wording, and "narrow the window" is weak advice for a chunk that's already pre-sliced. Consider generic wording ("trace query timed out") and emitting the narrow-window hint only on the non-chunked/aggregate path.

faiz9068 (commented):
Reviewed the fix against the backend query compiler (`tempo-api-for-traces/sqlquerygen/querygen.go`) — both failure modes are confirmed real: unknown operators (incl. `$exists`) hit `default: return "1=1"` (~L1633) and match every span (FDE-286); `$notnull` compiles to `(field != null)` (~L1623) which is never true in ClickHouse and matches nothing (ENG-1537).

**Strengths**
- `utils.PipelineHasAggregateStage` is a clean, nil-safe single source of truth; `logs.go` correctly refactored onto it. The aggregate-never-chunk guard fixes the duplicate group-by keys / wrong avg/median math.
- `rewriteExistsOperator` runs before validation and recurses through `$and/$or/$not`; the mid-range map mutation is safe. Good test coverage, incl. arg-less `$exists` rejection.
- `go build ./...` clean; traces/utils/logs tests pass.

**Ship-blocking finding that couldn't be inlined (file not in this diff):**
`internal/apm/databases.go:264` builds `{"$exists": ["attributes.db.system"]}` ("match any span that has db.system set"). Traced the path: `get_database_slow_queries` dispatches via `utils.MakeTracesJSONQueryAPI`, which does **not** route through `sanitizeTraceJSONQuery`/`rewriteExistsOperator`. So this `$exists` reaches the backend and compiles to `1=1` — matching *every* CLIENT/INTERNAL span, not just DB spans. It's the exact bug this PR documents, live in another path. Fix: `{"$neq": []any{"attributes.db.system", ""}}`, or hoist the rewrite into `MakeTracesJSONQueryAPI` so every trace-query producer is covered.

**Verdict: merge with fixes** — close the `$notnull` loop (see inline comments) and fix `databases.go:264`. The 408-wording and sibling-`$neq` notes are follow-ups.

internal/telemetry/traces/input_schema.go
faiz9068: Verified in `13394e9`: `$notnull` is removed from the `traceFilterFieldOperators` allow-list, and both `$exists` and `$notnull` are now rewritten to `{"$neq": [field, ""]}` via `rewriteBrokenExistenceOperators` (new `brokenExistenceOperators` list). New test covers the `$notnull` rewrite. Build + traces/utils/logs/apm tests all pass. Resolving.

internal/telemetry/traces/query_sanitize.go
faiz9068: Verified in `13394e9`: the validator error message no longer lists `$notnull` and now reads "... there is no $exists or $notnull operator." Consistent with the schema hint. Resolving.

internal/telemetry/traces/query_sanitize.go
faiz9068: Verified in `13394e9`: `rewriteBrokenExistenceOperatorsInMap` now folds sibling conditions into an `$and` instead of overwriting an existing `$neq`. New test "$exists alongside sibling $neq folds into $and" confirms both conditions survive. Resolving.

internal/telemetry/traces/traces.go
faiz9068: Verified in `13394e9`: the 408 message is now generic — "trace query timed out; narrow the time window or simplify the query and retry" — no longer aggregate-specific. Resolving.

faiz9068 (approved):

PR #201, opened 2026-08-05
fix(logs): create drop rules via otel_settings/drop API

prathamesh-sonpatki:
## Summary
- Route `add_drop_rule` through `POST /otel_settings/drop?region=...&cluster_id=...` instead of the legacy `PUT /logs_settings/routing` endpoint
- Send the nested `{ name, properties: { telemetry, filters, action } }` payload expected by the OTel settings API
- Return a clear HTTP error when the API rejects the request (instead of failing on JSON decode)

## Context
Alpha human-approval testing showed approvals working end-to-end, but drop rule creation failed with `PUT [405] /logs_settings/routing`. The API blocked write access to that route in favor of `/otel_settings/drop`.

## Test plan
- [x] `go test ./internal/telemetry/logs/... -run DropRule`
- [x] Approve an `add_drop_rule` action in alpha chat after deploying this build

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,12 +544,14 @@ Use `get_logs` for broad aggregate counts first; use `get_service_logs` only aft
 
 ### get_drop_rules
 
-No parameters.
+No parameters. Lists drop rules via `GET /otel_settings/drop?region=...`.
 
 ### add_drop_rule
 
 - `name` (string, required)
 - `filters` (array, required): Each filter: `key`, `value`, `operator` (`equals`/`not_equals`), `conjunction` (`and`).
+- Filter keys must use `attributes["key_name"]` or `resource.attributes["key_name"]` (required by last9-api).
+- Creates the rule via `POST /otel_settings/drop?region=...&cluster_id=...`.
 
 ### get_traces
 
--- a/README.md
+++ b/README.md
@@ -550,7 +550,7 @@ No parameters. Lists drop rules via `GET /otel_settings/drop?region=...`.
 
 - `name` (string, required)
 - `filters` (array, required): Each filter: `key`, `value`, `operator` (`equals`/`not_equals`), `conjunction` (`and`).
-- Filter keys must use `attributes["key_name"]` or `resource.attributes["key_name"]` (required by last9-api).
+- Filter keys must use `attributes["key_name"]` or `resource.attributes["key_name"]` (required by the Last9 API).
 - Creates the rule via `POST /otel_settings/drop?region=...&cluster_id=...`.
 
 ### get_traces
--- a/internal/constants/api.go
+++ b/internal/constants/api.go
@@ -28,8 +28,7 @@ const (
 	// Organization and configuration endpoints
 	EndpointDatasources          = "/datasources"
 	EndpointOAuthAccessToken     = "/api/v4/oauth/access_token"
-	EndpointLogsSettingsRouting = "/logs_settings/routing"
-	EndpointOTelSettingsDrop    = "/otel_settings/drop"
+	EndpointOTelSettingsDrop     = "/otel_settings/drop"
 	EndpointAlertRules           = "/alert-rules"
 	EndpointAlertsMonitor        = "/alerts/monitor"
 	EndpointEntitiesList         = "/entities/list"
--- a/internal/constants/api.go
+++ b/internal/constants/api.go
@@ -28,7 +28,8 @@ const (
 	// Organization and configuration endpoints
 	EndpointDatasources          = "/datasources"
 	EndpointOAuthAccessToken     = "/api/v4/oauth/access_token"
-	EndpointLogsSettingsRouting  = "/logs_settings/routing"
+	EndpointLogsSettingsRouting = "/logs_settings/routing"
+	EndpointOTelSettingsDrop    = "/otel_settings/drop"
 	EndpointAlertRules           = "/alert-rules"
 	EndpointAlertsMonitor        = "/alerts/monitor"
 	EndpointEntitiesList         = "/entities/list"
--- a/internal/models/logs.go
+++ b/internal/models/logs.go
@@ -1,6 +1,6 @@
 package models
 
-// DropRuleFilter represents a single filter condition in a drop rule
+// DropRuleFilter represents a single filter condition in a drop rule.
 type DropRuleFilter struct {
 	Key         string `json:"key"`
 	Value       string `json:"value"`
@@ -8,17 +8,31 @@ type DropRuleFilter struct {
 	Conjunction string `json:"conjunction"`
 }
 
-// DropRuleAction represents the action configuration for a drop rule
+// DropRuleAction represents the action configuration for a drop rule.
 type DropRuleAction struct {
 	Name        string                 `json:"name"`
 	Destination string                 `json:"destination"`
 	Properties  map[string]interface{} `json:"properties"`
 }
 
-// DropRule represents a complete drop rule configuration
+// DropRule is the flat shape returned by legacy GET /logs_settings/routing.
+// Do not use for writes; use OTelDropSettingCreateRequest with POST /otel_settings/drop.
 type DropRule struct {
 	Name      string           `json:"name"`
 	Telemetry string           `json:"telemetry"`
 	Filters   []DropRuleFilter `json:"filters"`
 	Action    DropRuleAction   `json:"action"`
+}
+
+// OTelDropSettingProperties is the nested properties payload for POST /otel_settings/drop.
+type OTelDropSettingProperties struct {
+	Telemetry string           `json:"telemetry"`
+	Filters   []DropRuleFilter `json:"filters"`
+	Action    DropRuleAction   `json:"action"`
+}
+
+// OTelDropSettingCreateRequest is the POST /otel_settings/drop request body.
+type OTelDropSettingCreateRequest struct {
+	Name       string                    `json:"name"`
+	Properties OTelDropSettingProperties `json:"properties"`
 }
--- a/internal/models/logs.go
+++ b/internal/models/logs.go
@@ -10,18 +10,9 @@ type DropRuleFilter struct {
 
 // DropRuleAction represents the action configuration for a drop rule.
 type DropRuleAction struct {
-	Name        string                 `json:"name"`
-	Destination string                 `json:"destination"`
-	Properties  map[string]interface{} `json:"properties"`
-}
-
-// DropRule is the flat shape returned by legacy GET /logs_settings/routing.
-// Do not use for writes; use OTelDropSettingCreateRequest with POST /otel_settings/drop.
-type DropRule struct {
-	Name      string           `json:"name"`
-	Telemetry string           `json:"telemetry"`
-	Filters   []DropRuleFilter `json:"filters"`
-	Action    DropRuleAction   `json:"action"`
+	Name        string            `json:"name"`
+	Destination string            `json:"destination"`
+	Properties  map[string]string `json:"properties"`
 }
 
 // OTelDropSettingProperties is the nested properties payload for POST /otel_settings/drop.
--- a/internal/telemetry/logs/drop_rule.go
+++ b/internal/telemetry/logs/drop_rule.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 
@@ -71,8 +72,8 @@ func NewGetDropRulesHandler(client *http.Client, cfg models.Config) func(context
 
 // DropRuleFilter represents a filter for drop rules
 type DropRuleFilter struct {
-	Key         string `json:"key" jsonschema:"The field to filter on (e.g. service level severity)"`
-	Value       string `json:"value" jsonschema:"The value to match against (e.g. test-service)"`
+	Key         string `json:"key" jsonschema:"The field to filter on (e.g. attributes[\"level\"], resource.attributes[\"service.name\"])"`
+	Value       string `json:"value" jsonschema:"The value to match against (e.g. debug)"`
 	Operator    string `json:"operator" jsonschema:"The operator to use for comparison (default: equals, options: equals, not_equals)"`
 	Conjunction string `json:"conjunction" jsonschema:"How to combine with other filters (default: and, options: and)"`
 }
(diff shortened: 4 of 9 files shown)


internal/utils/utils.go
```diff
@@ -463,3 +463,13 @@ func PopulateAPICfg(cfg *models.Config) error {
 
 	return nil
 }
+
+func validatePopulatedDatasourceCfg(cfg *models.Config) error {
+	if cfg.PrometheusReadURL == "" || cfg.PrometheusUsername == "" || cfg.PrometheusPassword == "" || cfg.Region == "" {
+		return errors.New("selected datasource missing required properties")
+	}
+	if cfg.ClusterID == "" {
```
faiz9068: Nit: no need for a separate branch here — `cluster_id` is a required datasource property like the others, so fold it into the existing check for consistency:

```go
if cfg.PrometheusReadURL == "" || cfg.PrometheusUsername == "" ||
    cfg.PrometheusPassword == "" || cfg.Region == "" || cfg.ClusterID == "" {
    return errors.New("selected datasource missing required properties")
}
```

Keeps one code path and one error message for the same class of failure.

internal/constants/api.go
```diff
@@ -28,7 +28,8 @@ const (
 	// Organization and configuration endpoints
 	EndpointDatasources          = "/datasources"
 	EndpointOAuthAccessToken     = "/api/v4/oauth/access_token"
-	EndpointLogsSettingsRouting  = "/logs_settings/routing"
+	EndpointLogsSettingsRouting = "/logs_settings/routing"
```
faiz9068: `EndpointLogsSettingsRouting` has zero references at this head — `git grep` finds only this definition. Please delete it.

Removing it also fixes the const-block alignment: `gofmt -l internal/constants/api.go` is dirty at this head and clean at the base commit, so this is a formatting regression introduced here.

internal/models/logs.go
```diff
@@ -1,24 +1,38 @@
...
-// DropRuleAction represents the action configuration for a drop rule
+// DropRuleAction represents the action configuration for a drop rule.
 type DropRuleAction struct {
 	Name        string                 `json:"name"`
 	Destination string                 `json:"destination"`
 	Properties  map[string]interface{} `json:"properties"`
 }
 
-// DropRule represents a complete drop rule configuration
+// DropRule is the flat shape returned by legacy GET /logs_settings/routing.
```
faiz9068: `models.DropRule` is also unreferenced at this head, and this new doc comment describes a read path that this same commit removed. Nothing ever decoded into the struct either — both handlers work with raw bytes. Leaving it in place implies a legacy fallback that doesn't exist, so please delete the type along with the comment.

For the record, `logs_settings` itself is **not** fully gone from this repo and shouldn't be — `internal/utils/log_index.go:111,135` still read `physical_indexes` and `rehydration`, and those GETs are still supported. It's only the drop-rule routing path that's dead.

internal/models/logs.go
```diff
@@ -1,24 +1,38 @@
...
 	Operator    string `json:"operator"`
 	Conjunction string `json:"conjunction"`
 }
 
-// DropRuleAction represents the action configuration for a drop rule
+// DropRuleAction represents the action configuration for a drop rule.
 type DropRuleAction struct {
 	Name        string                 `json:"name"`
 	Destination string                 `json:"destination"`
 	Properties  map[string]interface{} `json:"properties"`
```
faiz9068: `Properties map[string]interface{}` is wider than the API accepts — the field is a string map upstream. This works today only because the code always sends an empty map; the first non-string value put in here would fail at the API with a decode error rather than at compile time.

Narrowing to `map[string]string` makes the contract structurally enforced.

faiz9068 (commented):
Reviewed the full range. Build, vet, and `go test ./internal/...` all pass at this head — I ran them.

The core write-path fix is solid: `TestAddDropRuleHandlerSendsCorrectPayload` asserts the real wire contract (method, path, both query params, full nested body down to `action.destination`) rather than just a non-nil return, and moving the status check ahead of parsing in `readDropRuleAPIResponse` fixes exactly the failure mode that prompted this. The helper extraction in the last commit genuinely deduplicates the two handlers. Consistent `add_drop_rule:` / `get_drop_rules:` error prefixes are a nice improvement over the bare errors at base.

Moving the read path over as well is the right call — reading the same source you write to beats reading a mirror, and scoping that to drop rules only for now is the correct blast radius. Worth recording the observed response shape from a real call in the PR description, since the test fixture is hand-authored and `get_drop_rules` passes the body straight through to the model.

Remaining notes are cleanups rather than blockers: two now-dead symbols to delete (`EndpointLogsSettingsRouting`, `models.DropRule` — note `logs_settings` GETs are still live and still used by `log_index.go`, so this is cleanup, not removal), a gofmt regression in `internal/constants/api.go`, and the create-only / name-uniqueness change is worth surfacing in the tool description plus a mapped error so a retrying agent doesn't loop.

internal/telemetry/logs/drop_rule.go
```diff
@@ -19,184 +21,229 @@ import (
...
-}
-
-// AddDropRuleArgs represents the input arguments for adding drop rules
-type AddDropRuleArgs struct {
-	Name    string           `json:"name" jsonschema:"Name for the drop rule (e.g. test-service-drop-rule)"`
-	Filters []DropRuleFilter `json:"filters" jsonschema:"Array of filter conditions to match logs for dropping"`
-}
-
 // NewAddDropRuleHandler creates a handler for adding new drop rules for logs
 func NewAddDropRuleHandler(client *http.Client, cfg models.Config) func(context.Context, *mcp.CallToolRequest, AddDropRuleArgs) (*mcp.CallToolResult, any, error) {
```
faiz9068, line 185: **Behaviour change worth documenting.** The endpoint this now targets is create-only, and rule names are unique — where the previous write path tolerated a repeat. So calling `add_drop_rule` twice with the same `name` is no longer idempotent: the second call fails rather than being a no-op.

I checked the tool surface and nothing currently claims the old behaviour — `internal/prompts/descriptions/add_drop_rule.md` and the README both already say "create", so there's no stale doc to fix. The gap is that nothing tells the model the constraint exists either, so on a retry it gets an opaque failure with no idea what to do next.

Two things would close it:

1. Add a line to `internal/prompts/descriptions/add_drop_rule.md` — rule names must be unique; call `get_drop_rules` first if you might be re-adding one; this tool creates only, it does not update.
2. In `readDropRuleAPIResponse`, map the duplicate-name failure to a clear message ("a drop rule named X already exists") instead of passing the raw upstream error through.

Without (2) especially, a retrying agent is likely to loop.

internal/telemetry/logs/drop_rule.go
```diff
@@ -19,184 +21,229 @@ import (
...
+		if conjunction != "and" {
+			return nil, fmt.Errorf("invalid conjunction: %s. Must be: [and]", conjunction)
+		}
+		if f.Key == "" {
+			return nil, errors.New("key must be provided")
+		}
+		if f.Value == "" {
+			return nil, errors.New("value must be provided")
+		}
+		if !strings.HasPrefix(f.Key, "attributes[") && !strings.HasPrefix(f.Key, "resource.attributes[") {
```
faiz9068: Two things here.

**It's a tool-input contract change.** Keys that previously reached the API (e.g. a bare `service_name`, which the base test suite used) are now rejected client-side. That's fine as a decision, but it belongs under a `Changed` heading in the CHANGELOG rather than inside `Fixed` — callers passing bare keys will start seeing errors.

**It's weaker than the documented rule.** Prefix-only means `attributes[level]` (unquoted), `attributes[`, and `resource.attributes[]` all pass, while `add_drop_rule.md` states keys must use double quotes. Suggest validating the full form:

```go
var filterKeyRe = regexp.MustCompile(`^(resource\.)?attributes\["[^"]+"\]$`)
```

and adding a test case for unquoted `attributes[level]` being rejected — nothing covers that today.

internal/telemetry/logs/drop_rule.go
```diff
@@ -19,184 +21,229 @@ import (
...
+			Operator:    operator,
+			Conjunction: conjunction,
+		})
+	}
+
+	return converted, nil
+}
+
+func readDropRuleAPIResponse(resp *http.Response) ([]byte, error) {
+	respBody, err := io.ReadAll(resp.Body)
```
faiz9068, line 121: `io.ReadAll(resp.Body)` is unbounded, and the body now flows into the error string on non-2xx. An HTML error page from an intermediary would get embedded whole into the tool error and into the model's context.

`io.ReadAll(io.LimitReader(resp.Body, 64<<10))` plus truncation in the error message would cap it.

internal/telemetry/logs/drop_rule.go
```diff
@@ -19,184 +21,229 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response: %w", err)
+	}
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
+		return nil, fmt.Errorf("drop rule API request failed with status %d: %s", resp.StatusCode, string(respBody))
+	}
+	if !json.Valid(respBody) {
```
faiz9068, line 132: `json.Valid([]byte{})` is `false`, so a 2xx with an empty body (a `201`/`204` with no content) is reported as `drop rule API returned invalid JSON` **after the rule was successfully created**. The model will most likely retry — and then hit the duplicate-name failure noted above, turning one success into two confusing errors.

This flaw predates the PR, but since the stated goal here is "return a clear error instead of failing on JSON decode", it's worth closing: treat an empty body on 2xx as success and emit `{}`.

internal/telemetry/logs/drop_rule.go
```diff
@@ -19,184 +21,229 @@ import (
...
+		return nil, fmt.Errorf("drop rule API returned invalid JSON")
+	}
+	return respBody, nil
+}
+
+func dropRuleToolResult(cfg models.Config, responseBody []byte) (*mcp.CallToolResult, error) {
```
faiz9068: `dropRuleToolResult` returns an `error` it can never produce, so both call sites carry a dead `if err != nil`. Dropping the error return would simplify all three spots.

README.md
```diff
@@ -544,12 +544,14 @@ Use `get_logs` for broad aggregate counts first; use `get_service_logs` only aft
...
 ### get_drop_rules
 
-No parameters.
+No parameters. Lists drop rules via `GET /otel_settings/drop?region=...`.
 
 ### add_drop_rule
 
 - `name` (string, required)
 - `filters` (array, required): Each filter: `key`, `value`, `operator` (`equals`/`not_equals`), `conjunction` (`and`).
+- Filter keys must use `attributes["key_name"]` or `resource.attributes["key_name"]` (required by last9-api).
```
faiz9068: Nit: "(required by last9-api)" names an internal backend service in the user-facing README, and it's not information a user of this tool can act on. Suggest "required by the Last9 API".

internal/constants/api.go
faiz9068: Deleted in 11b4c1b. `gofmt -l internal/constants/api.go` is clean again.

internal/models/logs.go
faiz9068: Deleted in 11b4c1b, comment and all. Left the `logs_settings` GETs in `log_index.go` alone.

internal/models/logs.go
faiz9068: Narrowed to `map[string]string` in 11b4c1b — confirmed against the API, the action properties field is a string map there. Only ever sent empty, so no behaviour change.

internal/telemetry/logs/drop_rule.go
faiz9068: Dropped the error return in 11b4c1b; both call sites are now a single `return dropRuleToolResult(cfg, respBody), nil, nil`.

internal/telemetry/logs/drop_rule.go
faiz9068: Fixed in 11b4c1b: an empty body on 2xx now returns `{}` instead of `invalid JSON`. Test covers 200, 201 and 204.

README.md
faiz9068: Reworded to "the Last9 API" in 11b4c1b.

internal/utils/utils.go
faiz9068: Folded into the single check in 11b4c1b, with `cluster_id` alongside the other required properties. `utils_test.go` now asserts on the shared message.

internal/telemetry/logs/drop_rule.go
faiz9068: Went the other way on this one: the client-side filter key check is gone entirely in 11b4c1b, prefix version included. Key format is the API's contract to enforce, and mirroring it here just gives us two places to keep in sync. `convertDropRuleFilters` now only requires a non-empty key. Since nothing is rejected client-side any more, there's no input-contract change to record, and the CHANGELOG is back to a single line for the endpoint migration.

internal/telemetry/logs/drop_rule.go
faiz9068: Not doing this one — it adds a size-limiting layer for a response shape this API doesn't return. Left as a plain `io.ReadAll`.

internal/telemetry/logs/drop_rule.go
faiz9068: Split this one.

(1) is in 11b4c1b — `add_drop_rule.md` now carries a `Rule Names` line: names are unique, call `get_drop_rules` to see which are in use.

(2) isn't implementable from this side. The constraint is real, but a duplicate name doesn't come back as a 409 or with any distinguishing text — it surfaces as a generic 500. Keying on that would relabel every genuine server error as a duplicate name, which is worse than passing the error through. The API would need to return a distinct status for the conflict first; happy to file that separately.

faiz9068 (commented):
.

internal/telemetry/logs/drop_rule.go
```diff
@@ -19,184 +20,221 @@ import (
...
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response: %w", err)
+	}
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
+		return nil, fmt.Errorf("drop rule API request failed with status %d: %s", resp.StatusCode, string(respBody))
+	}
+	// json.Valid rejects an empty body, so a 201/204 would read as a failure.
+	if len(bytes.TrimSpace(respBody)) == 0 {
+		return []byte("{}"), nil
```
faiz9068: The empty-body fix is right for the create path, but `readDropRuleAPIResponse` is shared with `get_drop_rules`, and `{}` is the wrong empty value for a list.

Verified with a throwaway test against a server returning an empty `200`:

```
get_drop_rules empty-200 payload handed to the model => "{}"
```

So a tenant with no drop rules configured hands the model an object, where every populated response from that same tool is a JSON array. That's the kind of shape flip that gets a model to call `.length` on an object, or to read the call as having failed rather than as "no rules yet" — and it hits the tool's most common first call.

Cheapest fix is to let each caller decide its own empty value:

```go
func readDropRuleAPIResponse(resp *http.Response, emptyBody string) ([]byte, error) {
    ...
    // json.Valid rejects an empty body, so a 201/204 would read as a failure.
    if len(bytes.TrimSpace(respBody)) == 0 {
        return []byte(emptyBody), nil
    }
```

with `"[]"` from `NewGetDropRulesHandler` and `"{}"` from `NewAddDropRuleHandler`.

Worth extending `TestReadDropRuleAPIResponseEmptyBodyOnSuccess` to cover the list handler too — it currently only exercises `NewAddDropRuleHandler`, which is why this slipped through. Same loop over 200/201/204, asserting `[]`.

faiz9068 (commented):
One follow-up on the empty-body fix — inline.

internal/telemetry/logs/drop_rule.go
faiz9068: Fixed in 6c55980. `readDropRuleAPIResponse` takes the empty value from the caller — `"[]"` from `NewGetDropRulesHandler`, `"{}"` from `NewAddDropRuleHandler`.

`TestReadDropRuleAPIResponseEmptyBodyOnSuccess` now runs both handlers over 200/201/204 and asserts each one's own shape. Confirmed it catches the regression: forcing the list path back to `"{}"` fails with `status 201: expected [], got "{}"`.

One note on reachability — the list endpoint serialises an empty slice rather than sending a bodiless 200, so the branch shouldn't fire against the real API today. It was still the wrong default to leave sitting in a shared helper.

faiz9068 (approved):

PR #202, opened 2026-08-05
ENG-1543: Fix slow query trace filters to use bracket map notation

prathamesh-sonpatki:
## Summary
Fixes a correctness bug where `get_database_slow_queries` built trace pipeline filters with dot notation (`attributes.db.system`, `resource.attributes.deployment.environment`) that ClickHouse Map columns do not honor. Filters silently returned wrong/empty results when `db_system`, `host`, or `env` were set.

Extracts a shared pipeline builder (Option B) and routes all slow-query trace pipelines through `SanitizeTraceJSONPipeline` (Option C), which rewrites legacy dot notation to bracket syntax before forwarding.

## Changes
- `internal/apm/databases.go`: `buildDatabaseSlowQueryTracePipeline` + sanitizer pass before API call
- `internal/telemetry/traces/field_mapping.go`: `SpanAttributeField`, `ResourceAttributeField` helpers
- `internal/telemetry/traces/query_sanitize.go`: `SanitizeTraceJSONPipeline`, legacy dot-notation rewrite
- Tests: builder regression, handler httptest capture, sanitizer rewrite, helper unit tests

## Linear Issue
[ENG-1543](https://linear.app/last9/issue/ENG-1543/get-database-slow-queries-builds-trace-filters-with-dot-notation-on)

## Test Plan
- [x] `go test ./...` — 809 tests passing
- [x] TDD regression: pipeline tests fail on main (dot notation) / pass on fix
- [x] Coverage: `buildDatabaseSlowQueryTracePipeline` 94%, `SanitizeTraceJSONPipeline` 88%
- [ ] Manual: `get_database_slow_queries` with `db_system` + `env` returns trace-sourced slow queries

## Thermo Review
No medium/high findings. Low: handler e2e test only covers `DBSystem`-set path; optional follow-up for `$exists` + dot-notation interaction test.

### DIFF
--- a/internal/apm/databases.go
+++ b/internal/apm/databases.go
@@ -233,62 +233,6 @@ type SlowQuery struct {
 	RowsReturned int64  `json:"rows_returned,omitempty"`
 }
 
-func buildDatabaseSlowQueryTracePipeline(args GetDatabaseSlowQueriesArgs) ([]map[string]any, error) {
-	var conditions []any
-
-	conditions = append(conditions, map[string]any{
-		"$regex": []any{"SpanKind", "SPAN_KIND_CLIENT|SPAN_KIND_INTERNAL"},
-	})
-
-	if args.DBSystem != "" {
-		conditions = append(conditions, map[string]any{
-			"$eq": []any{traces.SpanAttributeField("db.system"), args.DBSystem},
-		})
-	} else {
-		conditions = append(conditions, map[string]any{
-			"$neq": []any{traces.SpanAttributeField("db.system"), ""},
-		})
-	}
-
-	if args.Host != "" {
-		conditions = append(conditions, map[string]any{
-			"$eq": []any{traces.SpanAttributeField("net.peer.name"), args.Host},
-		})
-	}
-
-	if args.ServiceName != "" {
-		conditions = append(conditions, map[string]any{
-			"$eq": []any{"ServiceName", args.ServiceName},
-		})
-	}
-
-	if args.Env != "" {
-		conditions = append(conditions, map[string]any{
-			"$eq": []any{traces.ResourceAttributeField("deployment.environment"), args.Env},
-		})
-	}
-
-	if args.MinDurationMs > 0 {
-		minDurationNs := int64(args.MinDurationMs * 1_000_000)
-		conditions = append(conditions, map[string]any{
-			"$gte": []any{"Duration", minDurationNs},
-		})
-	}
-
-	pipeline := []map[string]any{
-		{
-			"type":  "filter",
-			"query": map[string]any{"$and": conditions},
-		},
-	}
-
-	if err := traces.SanitizeTraceJSONQuery(pipeline); err != nil {
-		return nil, fmt.Errorf("invalid slow query trace pipeline: %w", err)
-	}
-
-	return pipeline, nil
-}
-
 func NewGetDatabaseSlowQueriesHandler(client *http.Client, cfg models.Config) func(context.Context, *mcp.CallToolRequest, GetDatabaseSlowQueriesArgs) (*mcp.CallToolResult, any, error) {
 	return func(ctx context.Context, req *mcp.CallToolRequest, args GetDatabaseSlowQueriesArgs) (*mcp.CallToolResult, any, error) {
 		startTime, endTime, err := resolveTimeRange(args.StartTimeISO, args.EndTimeISO, args.LookbackMinutes)
@@ -962,6 +906,62 @@ func queryPromInstantValue(ctx context.Context, client *http.Client, cfg models.
 }
 
 // --- helpers ---
+
+func buildDatabaseSlowQueryTracePipeline(args GetDatabaseSlowQueriesArgs) ([]map[string]any, error) {
+	var conditions []any
+
+	conditions = append(conditions, map[string]any{
+		"$regex": []any{"SpanKind", "SPAN_KIND_CLIENT|SPAN_KIND_INTERNAL"},
+	})
+
+	if args.DBSystem != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{traces.SpanAttributeField("db.system"), args.DBSystem},
+		})
+	} else {
+		conditions = append(conditions, map[string]any{
+			"$neq": []any{traces.SpanAttributeField("db.system"), ""},
+		})
+	}
+
+	if args.Host != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{traces.SpanAttributeField("net.peer.name"), args.Host},
+		})
+	}
+
+	if args.ServiceName != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{"ServiceName", args.ServiceName},
+		})
+	}
+
+	if args.Env != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{traces.ResourceAttributeField("deployment.environment"), args.Env},
+		})
+	}
+
+	if args.MinDurationMs > 0 {
+		minDurationNs := int64(args.MinDurationMs * 1_000_000)
+		conditions = append(conditions, map[string]any{
+			"$gte": []any{"Duration", minDurationNs},
+		})
+	}
+
+	pipeline := []map[string]any{
+		{
+			"type":  "filter",
+			"query": map[string]any{"$and": conditions},
+		},
+	}
+
+	if err := traces.SanitizeTraceJSONQuery(pipeline); err != nil {
+		return nil, fmt.Errorf("invalid slow query trace pipeline: %w", err)
+	}
+
+	return pipeline, nil
+}
 
 // fetchSlowQueryLogs queries the logs API for entries with attributes['slow_query']='true'
 // and extracts database-specific fields like plan_summary, docs_examined, etc.
--- a/internal/apm/databases.go
+++ b/internal/apm/databases.go
@@ -14,6 +14,7 @@ import (
 
 	"last9-mcp/internal/deeplink"
 	"last9-mcp/internal/models"
+	"last9-mcp/internal/telemetry/traces"
 	"last9-mcp/internal/utils"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
@@ -232,6 +233,62 @@ type SlowQuery struct {
 	RowsReturned int64  `json:"rows_returned,omitempty"`
 }
 
+func buildDatabaseSlowQueryTracePipeline(args GetDatabaseSlowQueriesArgs) ([]map[string]any, error) {
+	var conditions []any
+
+	conditions = append(conditions, map[string]any{
+		"$regex": []any{"SpanKind", "SPAN_KIND_CLIENT|SPAN_KIND_INTERNAL"},
+	})
+
+	if args.DBSystem != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{traces.SpanAttributeField("db.system"), args.DBSystem},
+		})
+	} else {
+		conditions = append(conditions, map[string]any{
+			"$neq": []any{traces.SpanAttributeField("db.system"), ""},
+		})
+	}
+
+	if args.Host != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{traces.SpanAttributeField("net.peer.name"), args.Host},
+		})
+	}
+
+	if args.ServiceName != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{"ServiceName", args.ServiceName},
+		})
+	}
+
+	if args.Env != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{traces.ResourceAttributeField("deployment.environment"), args.Env},
+		})
+	}
+
+	if args.MinDurationMs > 0 {
+		minDurationNs := int64(args.MinDurationMs * 1_000_000)
+		conditions = append(conditions, map[string]any{
+			"$gte": []any{"Duration", minDurationNs},
+		})
+	}
+
+	pipeline := []map[string]any{
+		{
+			"type":  "filter",
+			"query": map[string]any{"$and": conditions},
+		},
+	}
+
+	if err := traces.SanitizeTraceJSONPipeline(pipeline); err != nil {
+		return nil, fmt.Errorf("invalid slow query trace pipeline: %w", err)
+	}
+
+	return pipeline, nil
+}
+
 func NewGetDatabaseSlowQueriesHandler(client *http.Client, cfg models.Config) func(context.Context, *mcp.CallToolRequest, GetDatabaseSlowQueriesArgs) (*mcp.CallToolResult, any, error) {
 	return func(ctx context.Context, req *mcp.CallToolRequest, args GetDatabaseSlowQueriesArgs) (*mcp.CallToolResult, any, error) {
 		startTime, endTime, err := resolveTimeRange(args.StartTimeISO, args.EndTimeISO, args.LookbackMinutes)
@@ -244,59 +301,9 @@ func NewGetDatabaseSlowQueriesHandler(client *http.Client, cfg models.Config) fu
 			limit = 20
 		}
 
-		// Build trace query pipeline filters.
-		// The traces API uses dot-notation for nested attributes:
-		//   - span attributes: "attributes.db.system"
-		//   - resource attributes: "resource.attributes.deployment.environment"
-		var conditions []any
-
-		// Filter by SPAN_KIND_CLIENT or SPAN_KIND_INTERNAL (DB operations)
-		conditions = append(conditions, map[string]any{
-			"$regex": []any{"SpanKind", "SPAN_KIND_CLIENT|SPAN_KIND_INTERNAL"},
-		})
-
-		if args.DBSystem != "" {
-			conditions = append(conditions, map[string]any{
-				"$eq": []any{"attributes.db.system", args.DBSystem},
-			})
-		} else {
-			// No specific db_system — match any span that has db.system set
-			conditions = append(conditions, map[string]any{
-				"$neq": []any{"attributes['db.system']", ""},
-			})
-		}
-
-		if args.Host != "" {
-			conditions = append(conditions, map[string]any{
-				"$eq": []any{"attributes.net.peer.name", args.Host},
-			})
-		}
-
-		if args.ServiceName != "" {
-			conditions = append(conditions, map[string]any{
-				"$eq": []any{"ServiceName", args.ServiceName},
-			})
-		}
-
-		if args.Env != "" {
-			conditions = append(conditions, map[string]any{
-				"$eq": []any{"resource.attributes.deployment.environment", args.Env},
-			})
-		}
-
-		if args.MinDurationMs > 0 {
-			// Duration in traces is in nanoseconds
-			minDurationNs := int64(args.MinDurationMs * 1_000_000)
-			conditions = append(conditions, map[string]any{
-				"$gte": []any{"Duration", minDurationNs},
-			})
-		}
-
-		pipeline := []map[string]any{
-			{
-				"type":  "filter",
-				"query": map[string]any{"$and": conditions},
-			},
+		pipeline, err := buildDatabaseSlowQueryTracePipeline(args)
+		if err != nil {
+			return nil, nil, err
 		}
 
 		// Use milliseconds for traces API
--- a/internal/apm/databases.go
+++ b/internal/apm/databases.go
@@ -282,7 +282,7 @@ func buildDatabaseSlowQueryTracePipeline(args GetDatabaseSlowQueriesArgs) ([]map
 		},
 	}
 
-	if err := traces.SanitizeTraceJSONPipeline(pipeline); err != nil {
+	if err := traces.SanitizeTraceJSONQuery(pipeline); err != nil {
 		return nil, fmt.Errorf("invalid slow query trace pipeline: %w", err)
 	}
 
--- a/internal/apm/databases_test.go
+++ b/internal/apm/databases_test.go
@@ -409,11 +409,21 @@ func TestBuildDatabaseSlowQueryTracePipeline_UsesBracketNotation(t *testing.T) {
 }
 
 func TestGetDatabaseSlowQueriesHandler_TracePipelineUsesBracketNotation(t *testing.T) {
+	// Assertions run on the server goroutine: Fatalf would Goexit there, and a
+	// request that never arrives would pass the test vacuously.
+	var mu sync.Mutex
+	reached := false
+
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if strings.Contains(r.URL.Path, "/cat/api/traces/v2/query_range/json") {
+			mu.Lock()
+			reached = true
+			mu.Unlock()
+
 			body, err := io.ReadAll(r.Body)
 			if err != nil {
-				t.Fatalf("failed to read request body: %v", err)
+				t.Errorf("failed to read request body: %v", err)
+				return
 			}
 
 			var req struct {
@@ -422,26 +432,29 @@ func TestGetDatabaseSlowQueriesHandler_TracePipelineUsesBracketNotation(t *testi
 				} `json:"pipeline"`
 			}
 			if err := json.Unmarshal(body, &req); err != nil {
-				t.Fatalf("failed to unmarshal request body: %v", err)
+				t.Errorf("failed to unmarshal request body: %v", err)
+				return
 			}
 			if len(req.Pipeline) != 1 {
-				t.Fatalf("expected exactly one pipeline stage, got %d", len(req.Pipeline))
+				t.Errorf("expected exactly one pipeline stage, got %d", len(req.Pipeline))
+				return
 			}
 
 			rawQuery, err := json.Marshal(req.Pipeline[0].Query)
 			if err != nil {
-				t.Fatalf("failed to marshal stage query: %v", err)
+				t.Errorf("failed to marshal stage query: %v", err)
+				return
 			}
 			query := string(rawQuery)
 
 			if !strings.Contains(query, `attributes['db.system']`) {
-				t.Fatalf("expected attributes['db.system'] in pipeline, got %s", query)
+				t.Errorf("expected attributes['db.system'] in pipeline, got %s", query)
 			}
 			if !strings.Contains(query, `attributes['net.peer.name']`) {
-				t.Fatalf("expected attributes['net.peer.name'] in pipeline, got %s", query)
+				t.Errorf("expected attributes['net.peer.name'] in pipeline, got %s", query)
 			}
 			if !strings.Contains(query, `resources['deployment.environment']`) {
-				t.Fatalf("expected resources['deployment.environment'] in pipeline, got %s", query)
+				t.Errorf("expected resources['deployment.environment'] in pipeline, got %s", query)
 			}
 			for _, bad := range []string{
 				"attributes.db.system",
@@ -449,7 +462,7 @@ func TestGetDatabaseSlowQueriesHandler_TracePipelineUsesBracketNotation(t *testi
 				"resource.attributes.deployment.environment",
 			} {
 				if strings.Contains(query, bad) {
-					t.Fatalf("pipeline should not include legacy dot notation %q, got %s", bad, query)
+					t.Errorf("pipeline should not include legacy dot notation %q, got %s", bad, query)
 				}
 			}
 
@@ -476,6 +489,12 @@ func TestGetDatabaseSlowQueriesHandler_TracePipelineUsesBracketNotation(t *testi
 	})
 	if err != nil {
 		t.Fatalf("handler returned error: %v", err)
+	}
+
+	mu.Lock()
+	defer mu.Unlock()
+	if !reached {
+		t.Fatal("traces query_range endpoint was never called — pipeline assertions never ran")
 	}
 }
 
--- a/internal/apm/databases_test.go
+++ b/internal/apm/databases_test.go
@@ -339,6 +339,146 @@ func TestGetDatabaseSlowQueriesHandler(t *testing.T) {
 	}
 }
 
+func TestBuildDatabaseSlowQueryTracePipeline_UsesBracketNotation(t *testing.T) {
+	t.Run("with db_system host env service and min duration", func(t *testing.T) {
+		pipeline, err := buildDatabaseSlowQueryTracePipeline(GetDatabaseSlowQueriesArgs{
+			DBSystem:      "postgresql",
+			Host:          "db.example.com",
+			ServiceName:   "checkout",
+			Env:           "prod",
+			MinDurationMs: 100,
+		})
+		if err != nil {
+			t.Fatalf("buildDatabaseSlowQueryTracePipeline() error = %v", err)
+		}
+
+		rawQuery, err := json.Marshal(pipeline[0]["query"])
+		if err != nil {
+			t.Fatalf("failed to marshal pipeline query: %v", err)
+		}
+		query := string(rawQuery)
+
+		for _, want := range []string{
+			`attributes['db.system']`,
+			`attributes['net.peer.name']`,
+			`resources['deployment.environment']`,
+			"postgresql",
+			"db.example.com",
+			"checkout",
+			"prod",
+			"100000000",
+		} {
+			if !strings.Contains(query, want) {
+				t.Fatalf("expected pipeline to include %q, got %s", want, query)
+			}
+		}
+
+		for _, bad := range []string{
+			"attributes.db.system",
+			"attributes.net.peer.name",
+			"resource.attributes.deployment.environment",
+		} {
+			if strings.Contains(query, bad) {
+				t.Fatalf("pipeline should not include legacy dot notation %q, got %s", bad, query)
+			}
+		}
+	})
+
+	t.Run("without db_system matches any span with db.system set", func(t *testing.T) {
+		pipeline, err := buildDatabaseSlowQueryTracePipeline(GetDatabaseSlowQueriesArgs{
+			LookbackMinutes: 30,
+		})
+		if err != nil {
+			t.Fatalf("buildDatabaseSlowQueryTracePipeline() error = %v", err)
+		}
+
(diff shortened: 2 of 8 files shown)


faiz9068 (commented):
Reviewed the diff plus the surrounding code in `databases.go`, `field_mapping.go` and `query_sanitize.go`. The core fix is correct and I verified it against live data — replaying the exact pipeline `buildDatabaseSlowQueryTracePipeline` now emits returns real spans, and the pre-fix dot-notation form reproduces `400 {"detail":"parse error: invalid JSON pipeline"}` on the currently released build. Worth noting the failure is a hard 400 on *every* filtered call, not the "silently wrong/empty results" the issue describes — so the user-visible impact is bigger than the ticket implies.

`go build ./...` and `go vet ./...` are clean; the only failing tests are the `internal/dashboards` integration tests that need `LAST9_DELETE_TOKEN`, i.e. unrelated to this branch.

### What's good

- `resources['deployment.environment']` (the issue's open question) is corroborated by three existing places — `internal/telemetry/traces/service_traces.go:176`, `internal/prompts/references/tracejson.md:32`, and `get_traces_base.md:13` — rather than guessed.
- Delegating to `enrichAttribute` rather than hardcoding a second mapping pays off in a non-obvious way: `ResourceAttributeField("service.name")` falls through to `ServiceName` instead of `resources['service.name']`, which a hand-rolled prefix would have gotten wrong.
- Sanitizer ordering is right — `rewriteBrokenExistenceOperators` runs before the new pass, so the `$neq`s it synthesizes still get their fields rewritten. Swapping those two lines would leave a silent hole.
- `databases.go:249` also resolves a pre-existing inconsistency where the `$neq` branch already used brackets while `$eq` did not.
- Side effect worth calling out: `databases.go:402` feeds the same sanitized pipeline into `BuildTracesLink`, so the returned deep link was broken before this too. Nothing pins that in a test, though — a future refactor that builds the link from a pre-sanitize copy would regress it silently.

### Should fix

**1. The sanitizer closes only two of the three map columns.** `normalizeTraceFilterField` (`query_sanitize.go:326-339`) passes through `attributes['`, `resources['` and `events['`, so all three columns were clearly in mind — but it only *rewrites* `resource.attributes.` and `attributes.`. These fall through untouched and hit the same 400 this PR exists to remove:

- `events.exception.type`
- `resources.deployment.environment` (plural + dotted)
- `resource.deployment.environment` (no `attributes.` segment)

`validateFieldSyntax` doesn't reject them either. This matters more than it looks: the handler's fields are hardcoded correct now, so the sanitizer is the only line of defense for model-authored `get_traces` pipelines, which is the high-volume path. `enrichAttribute` already maps `event_*`, so an `EventAttributeField` sibling is ~3 lines.

**2. The new handler test can pass while asserting nothing.** In `TestGetDatabaseSlowQueriesHandler_TracePipelineUsesBracketNotation` (`databases_test.go:390-455`) every assertion sits inside the `httptest` closure behind the `/cat/api/traces/v2/query_range/json` path check. If that branch is never entered, the fallback `w.WriteHeader(http.StatusOK)` returns an empty result, the handler returns no error, and the test goes green having verified nothing. Suggest a guarded `reached` flag asserted after the call.

Also, `t.Fatalf` inside that closure runs on the server's goroutine, not the test's. `FailNow`/`Fatalf` are only valid from the test goroutine — here it calls `runtime.Goexit` on the server goroutine, aborting the request mid-flight and reporting confusingly instead of failing cleanly. `t.Errorf` + `return` is the fix.

(The pure unit test `TestBuildDatabaseSlowQueryTracePipeline_UsesBracketNotation` is the right shape — both branches, positive *and* negative assertions.)

**3. Pre-existing, probably a separate issue: the two fetch tiers disagree on which filters exist.** `fetchSlowQueryLogs` (`databases.go:969-1003`) applies `DBSystem`, `Env` and `ServiceName` but ignores `Host` and `MinDurationMs`, both of which the traces tier honors (`databases.go:253-257`, `271-276`). Since the merged results are sorted by duration desc and truncated to `limit` (`databases.go:358-368`), an unfiltered high-duration log row from an unrelated host can evict a trace row that *did* match the filters. Pass `host` and you get trace rows from that host mixed with log rows from everywhere, indistinguishable in one array. Not introduced here, but this PR puts the two tiers side by side, so it seems worth filing.

The tiers also differ on `db.system` — traces check span attributes only, logs check `attributes` OR `resources` (`databases.go:978-984`). That asymmetry looks deliberate (`db.system` is a span attribute per semconv; the log-side `$or` hedges against varied log shapers) but a one-line comment would stop the next reader from "fixing" one to match the other.

### Minor

- `query_sanitize.go:287-299` — the type-conversion body is dead code. `any` is an alias for `interface{}`, so `[]map[string]any` and `[]map[string]interface{}` are the identical type, not merely convertible; the body reduces to `return sanitizeTraceJSONQuery(stages)`. As written it also implies a deep copy is happening, which would contradict the documented "applied in place" contract.
- `databases.go:285-289` — this error can't currently fire (all field refs are helper-generated or literal, and only `args[0]` is validated). Fine as future insurance, just noting the `error` in the signature is for an unreachable branch.
- `field_mapping.go:98-100` — `ResourceAttributeField("resource_foo")` yields `resources['resource_foo']`. Unreachable from current call sites, but it's exported now; a `strings.TrimPrefix` guard or a doc note that the key must be unprefixed would save the next caller.
- `service_traces.go:503-508` reads both `deployment.environment` and `deployment_environment`, implying some orgs emit the underscore form — the new `env` filter only matches the dotted one. Consistent with `service_traces.go:176` so not a regression, but if the underscore form is real in production both filters want an `$or`.
- `net.peer.name` is legacy semconv; newer SDKs emit `server.address`. Consistent with the metrics tier's `net_peer_name` label so correctly out of scope here, but the `host` filter will silently miss spans from upgraded SDKs. Follow-up ticket.
- `rewriteBrokenExistenceOperators`, `rewriteLegacyDotNotationFields` and `validateFilterFields` are now three near-identical tree walkers. Not worth abstracting at three, but a fourth should trigger a shared `walkFilterFields(query, func(field string) string)`.
- No `CHANGELOG.md` entry — `## [Unreleased]` has `### Added` but no `### Fixed`, and this is a user-visible fix to a tool that was returning 400. (#195 skipped it too, so soft convention at best.)

### Summary

Fix is sound and the factoring is an improvement. I'd fold **#1** in here since it's what turns this from "one handler fixed" into "bug class closed", and **#2** because a vacuously-passing test is worse than no test. **#3** reads like its own issue. Leaving this as a comment rather than an approval.

internal/telemetry/traces/query_sanitize.go
```diff
@@ -281,6 +282,62 @@ func validateFieldSyntax(field, path string) error {
...
+		strings.HasPrefix(field, `events['`) {
+		return field
+	}
+	if strings.HasPrefix(field, "resource.attributes.") {
+		return ResourceAttributeField(field[len("resource.attributes."):])
+	}
+	if strings.HasPrefix(field, "attributes.") {
+		return SpanAttributeField(field[len("attributes."):])
+	}
+	return field
```
faiz9068, line 340: **Should fix — this closes only two of the three map columns.**

The pass-through list above covers `attributes['`, `resources['` and `events['`, so all three columns were clearly in mind. But only `resource.attributes.` and `attributes.` get *rewritten*. These fall through untouched and hit the same 400 this PR exists to remove:

- `events.exception.type`
- `resources.deployment.environment` (plural + dotted)
- `resource.deployment.environment` (no `attributes.` segment)

`validateFieldSyntax` doesn't reject them either — it only catches `ResourceAttributes.`, `SpanAttributes.`, double-quoted brackets, and flat `resource_`/`event_` prefixes. So a model emitting `events.exception.type` gets precisely the opaque upstream `parse error: invalid JSON pipeline` with nothing to self-correct from.

This matters more here than in the handler: `buildDatabaseSlowQueryTracePipeline`'s fields are hardcoded correct now, so this function is the only line of defense for model-authored `get_traces` pipelines — the high-volume path. `enrichAttribute` already maps `event_*`, so an `EventAttributeField` sibling is ~3 lines.

internal/telemetry/traces/query_sanitize.go
```diff
@@ -281,6 +282,62 @@ func validateFieldSyntax(field, path string) error {
...
+	converted := make([]map[string]interface{}, len(stages))
+	for i, stage := range stages {
+		converted[i] = stage
+	}
+	if err := sanitizeTraceJSONQuery(converted); err != nil {
+		return err
+	}
+	for i := range stages {
+		stages[i] = converted[i]
+	}
```
faiz9068: **Minor — this is dead code.** `any` is an alias for `interface{}`, so `[]map[string]any` and `[]map[string]interface{}` are the *identical* type, not merely convertible. Both loops and `converted` can go; the body reduces to `return sanitizeTraceJSONQuery(stages)`.

Verified it compiles:

```go
func f(s []map[string]interface{}) int { return len(s) }
func main() { var g []map[string]any; _ = f(g) }
```

Worth removing rather than leaving: as written it implies a deep copy is happening, which would contradict the "applied in place" contract in the doc comment — misleading in the one direction that matters.

internal/telemetry/traces/field_mapping.go
```diff
@@ -89,6 +89,16 @@ func enrichAttribute(raw string) TraceAttribute {
 	}
 }
 
+// SpanAttributeField returns the tracejson filter field for a span attribute key.
+func SpanAttributeField(key string) string {
+	return enrichAttribute(key).FilterField
+}
+
+// ResourceAttributeField returns the tracejson filter field for a resource attribute key.
+func ResourceAttributeField(key string) string {
```
faiz9068, line 100: **Minor** — `ResourceAttributeField("resource_foo")` yields `resources['resource_foo']`. Not reachable from current call sites, but this is exported now, so a `strings.TrimPrefix(key, "resource_")` guard — or a doc note that the key must be unprefixed — would save the next caller a confusing empty result set.

Separately, nice property of routing through `enrichAttribute` here: `ResourceAttributeField("service.name")` correctly returns `ServiceName` rather than `resources['service.name']`, so the sanitizer gets `resource.attributes.service.name` right for free.

internal/apm/databases_test.go
```diff
@@ -339,6 +339,146 @@ func TestGetDatabaseSlowQueriesHandler(t *testing.T) {
...
+		}
+		if strings.Contains(query, "attributes.db.system") {
+			t.Fatalf("pipeline should not include legacy dot notation, got %s", query)
+		}
+	})
+}
+
+func TestGetDatabaseSlowQueriesHandler_TracePipelineUsesBracketNotation(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if strings.Contains(r.URL.Path, "/cat/api/traces/v2/query_range/json") {
```
faiz9068, line 418: **Should fix — this test can pass while asserting nothing.**

Every assertion in this test lives inside the closure, behind this path check. If the branch is never entered — path constant changes, or the traces call errors out before dispatch — the fallback `w.WriteHeader(http.StatusOK)` at the bottom returns an empty result, the handler returns no error, and the test goes green having verified nothing at all.

Suggest a `reached` flag (guarded, since this runs on the server's goroutine) asserted after the handler call, so the test fails loudly if the request never arrives.

internal/apm/databases_test.go
```diff
@@ -339,6 +339,146 @@ func TestGetDatabaseSlowQueriesHandler(t *testing.T) {
...
+			body, err := io.ReadAll(r.Body)
+			if err != nil {
+				t.Fatalf("failed to read request body: %v", err)
```
faiz9068: **Should fix — `t.Fatalf` from the wrong goroutine.** This closure runs on the `httptest` server's goroutine, not the test's. `testing` documents `FailNow`/`Fatalf` as callable only from the goroutine running the test; here it calls `runtime.Goexit` on the server goroutine, which aborts the request mid-flight and reports confusingly instead of failing cleanly. Applies to every `t.Fatalf` in this closure.

`t.Errorf` + `return` is the fix. Note the pre-existing stub at `databases_test.go:240-276` sidesteps this by keeping assertions out of the closure entirely.

FWIW `TestBuildDatabaseSlowQueryTracePipeline_UsesBracketNotation` above is the right shape — both branches, positive *and* negative assertions, no goroutine involved.

internal/apm/databases.go
```diff
@@ -232,71 +233,77 @@ type SlowQuery struct {
...
+	pipeline := []map[string]any{
+		{
+			"type":  "filter",
+			"query": map[string]any{"$and": conditions},
+		},
+	}
+
+	if err := traces.SanitizeTraceJSONPipeline(pipeline); err != nil {
+		return nil, fmt.Errorf("invalid slow query trace pipeline: %w", err)
+	}
```
faiz9068: **Minor, no action needed** — just flagging that this branch can't currently fire: every field ref in the pipeline is helper-generated or a literal top-level field, and only `args[0]` is validated (user input lands in `args[1]`, unvalidated by design). Fine as insurance against a future edit — worth knowing the `error` added to this function's signature is for an unreachable path.

internal/apm/databases.go
```diff
@@ -232,71 +233,77 @@ type SlowQuery struct {
...
 
-		if args.Host != "" {
-			conditions = append(conditions, map[string]any{
-				"$eq": []any{"attributes.net.peer.name", args.Host},
-			})
-		}
+	if args.Env != "" {
+		conditions = append(conditions, map[string]any{
+			"$eq": []any{traces.ResourceAttributeField("deployment.environment"), args.Env},
```
faiz9068: **Minor / follow-up** — `service_traces.go:503-508` reads *both* `deployment.environment` and `deployment_environment` from span resource attributes, which implies some orgs emit the underscore form. This filter only matches the dotted one, so `env` will silently return zero traces for those orgs.

Consistent with `service_traces.go:176` so not a regression from this PR. But if the underscore form is real in production, both filters want an `$or`.

faiz9068 (commented):
Reviewed the diff plus the surrounding code in `databases.go`, `field_mapping.go` and `query_sanitize.go`. Detailed notes are inline; summary here, plus two items that fall outside the diff so I couldn't anchor them to a line.

The core fix is correct and I verified it against live data — replaying the exact pipeline `buildDatabaseSlowQueryTracePipeline` now emits returns real spans, and the pre-fix dot-notation form reproduces `400 {"detail":"parse error: invalid JSON pipeline"}` on the currently released build. Worth noting the failure is a hard 400 on *every* filtered call, not the "silently wrong/empty results" the issue describes — the user-visible impact is bigger than the ticket implies.

`go build ./...` and `go vet ./...` are clean; the only failing tests are the `internal/dashboards` integration tests that need `LAST9_DELETE_TOKEN`, i.e. unrelated to this branch.

### What's good

- `resources['deployment.environment']` (the issue's open question) is corroborated by three existing places — `internal/telemetry/traces/service_traces.go:176`, `internal/prompts/references/tracejson.md:32`, and `get_traces_base.md:13` — rather than guessed.
- Delegating to `enrichAttribute` rather than hardcoding a second mapping pays off in a non-obvious way: `ResourceAttributeField("service.name")` falls through to `ServiceName` instead of `resources['service.name']`, which a hand-rolled prefix would have gotten wrong.
- Sanitizer ordering is right — `rewriteBrokenExistenceOperators` runs before the new pass, so the `$neq`s it synthesizes still get their fields rewritten. Swapping those two lines would leave a silent hole.
- `buildDatabaseSlowQueryTracePipeline` also resolves a pre-existing inconsistency where the `$neq` branch already used brackets while `$eq` did not.

### Outside the diff

**Untested side effect worth pinning.** `databases.go:402` feeds the same sanitized pipeline into `BuildTracesLink`, so the deep link this tool returns was broken before this PR too — a real second fix. Nothing pins it, so a future refactor that builds the link from a pre-sanitize copy would regress it silently.

**Pre-existing, probably its own issue: the two fetch tiers disagree on which filters exist.** `fetchSlowQueryLogs` (`databases.go:969-1003`) applies `DBSystem`, `Env` and `ServiceName` but ignores `Host` and `MinDurationMs`, both of which the traces tier honors. Since the merged results are sorted by duration desc and truncated to `limit` (`databases.go:358-368`), an unfiltered high-duration log row from an unrelated host can evict a trace row that *did* match the filters. Pass `host` and you get trace rows from that host mixed with log rows from everywhere, indistinguishable in one array.

The tiers also differ on `db.system` — traces check span attributes only, logs check `attributes` OR `resources` (`databases.go:978-984`). That asymmetry looks deliberate (`db.system` is a span attribute per semconv; the log-side `$or` hedges against varied log shapers) but a one-line comment would stop the next reader from "fixing" one to match the other.

**Minor:** no `CHANGELOG.md` entry — `## [Unreleased]` has `### Added` but no `### Fixed`, and this is a user-visible fix to a tool that was returning 400. (#195 skipped it too, so soft convention at best.)

### Summary

Fix is sound and the factoring is an improvement. I'd fold the `events.`/`resources.` sanitizer gap in here since it's what turns this from "one handler fixed" into "bug class closed", and the test-assertion issue because a vacuously-passing test is worse than no test. Leaving this as a comment rather than an approval.

internal/apm/databases.go
```diff
@@ -232,71 +233,77 @@ type SlowQuery struct {
...
+		conditions = append(conditions, map[string]any{
+			"$neq": []any{traces.SpanAttributeField("db.system"), ""},
+		})
+	}
 
-		// Filter by SPAN_KIND_CLIENT or SPAN_KIND_INTERNAL (DB operations)
+	if args.Host != "" {
 		conditions = append(conditions, map[string]any{
-			"$regex": []any{"SpanKind", "SPAN_KIND_CLIENT|SPAN_KIND_INTERNAL"},
+			"$eq": []any{traces.SpanAttributeField("net.peer.name"), args.Host},
```
faiz9068: **Minor / follow-up, out of scope here** — `net.peer.name` is legacy semconv; newer OTel SDKs emit `server.address` for the same concept. Consistent with the metrics tier's `net_peer_name` label (`databases.go:59`), so correctly unchanged in this PR — but the `host` filter will silently miss spans from recently-upgraded SDKs. Probably its own ticket.

internal/telemetry/traces/query_sanitize.go
faiz9068: Fixed in d3364b3. \`normalizeTraceFilterField\` now walks an ordered prefix table — \`resource.attributes.\`, \`resources.\`, \`resource.\`, \`attributes.\`, \`events.\` — longest first so \`resource.attributes.foo\` still beats \`resource.\`. Added \`EventAttributeField\` alongside the existing helpers.

Tests cover all three you listed plus \`resource.service.name\` → \`ServiceName\` (the \`enrichAttribute\` fallthrough), and a pass-through case asserting already-bracketed fields and top-level names are untouched.

internal/telemetry/traces/query_sanitize.go
faiz9068: Fixed, and went one further — the wrapper is gone entirely rather than reduced to a one-line delegate. \`sanitizeTraceJSONQuery\` is exported as \`SanitizeTraceJSONQuery\` and \`databases.go\` calls it directly, so there is no second entry point to keep in sync. Dropped \`TestSanitizeTraceJSONPipeline_RewritesLegacyDotNotationInPlace\` with it: it only existed to cover the wrapper and duplicated the \`attributes.db.system\` subtest exactly.

internal/telemetry/traces/field_mapping.go
faiz9068: Fixed in code rather than by doc note — both helpers now \`TrimPrefix\` the flat form, so \`resource_deployment.environment\` and \`deployment.environment\` both yield \`resources['deployment.environment']\`. Same for \`EventAttributeField\` and \`event_\`.

I had initially gone with the doc note, reasoning that a genuine attribute literally named \`resource_foo\` would be mangled by the guard. That is the wrong tradeoff: \`resource_foo\` is the flat API spelling of resource attribute \`foo\`, so \`TrimPrefix\` is what a caller passing it actually means. Both spellings are pinned in \`TestResourceAttributeField\`/\`TestEventAttributeField\`.

internal/apm/databases_test.go
faiz9068: Fixed. Added a mutex-guarded \`reached\` flag set on entry to the branch and asserted after the handler call, so a never-dispatched request fails loudly instead of going green.

internal/apm/databases_test.go
faiz9068: Fixed. Every \`t.Fatalf\` in the closure is now \`t.Errorf\` + \`return\`. Comment on the closure records both hazards (Goexit on the server goroutine, and the vacuous pass) so the next edit does not reintroduce either.

internal/apm/databases.go
faiz9068: Acknowledged, no change. Left the \`error\` in the signature as insurance for a future edit that introduces a caller-supplied field ref.

internal/apm/databases.go
faiz9068: Agreed, follow-up. Not touched here — will file separately alongside the \`net.peer.name\` item.

internal/apm/databases.go
faiz9068: Agreed, out of scope. Filing with the \`deployment_environment\` item as one semconv-drift ticket for the \`host\`/\`env\` filters.

internal/telemetry/traces/query_sanitize.go
```diff
@@ -281,6 +282,55 @@ func validateFieldSyntax(field, path string) error {
...
+		{"resource.attributes.", ResourceAttributeField},
+		{"resources.", ResourceAttributeField},
+		{"resource.", ResourceAttributeField},
+		{"attributes.", SpanAttributeField},
+		{"events.", EventAttributeField},
+	} {
+		if strings.HasPrefix(field, p.prefix) {
+			return p.to(field[len(p.prefix):])
+		}
+	}
```
faiz9068, line 339: The prefix-table rewrite is a good shape — ordered longest-first, and the comment explains why the order matters. Verified by execution that all the previously-escaping forms now land on the right map column.

**Minor: degenerate prefix-only inputs get rewritten into the *wrong* column.**

```
"events."              → attributes['event_']
"resources."           → attributes['resource_']
"resource."            → attributes['resource_']
"resource.attributes." → attributes['resource_']
"attributes."          → attributes['']
```

Cause is specific: with an empty remainder, `EventAttributeField("")` builds the string `"event_"`, and `enrichAttribute` guards its event branch with `len(raw) > 6` (`field_mapping.go:49`) — the exact-length string misses it and falls through to Priority 6, which wraps anything unrecognized in `attributes[...]`. Same story for `"resource_"`.

Only reachable from nonsense input, so the practical impact is a confusing 400 instead of a useful sanitizer error. Flagging it because this is the one place the expanded rewrite can pick the wrong map column, which is the failure mode this PR exists to eliminate.

Fix either way: skip the rewrite when the stripped remainder is empty (return `field` untouched and let `validateFieldSyntax` complain), or relax the `enrichAttribute` length guards to `>=`.

internal/telemetry/traces/query_sanitize.go
```diff
@@ -281,6 +282,55 @@ func validateFieldSyntax(field, path string) error {
...
+	// Longest prefix first: resource.attributes. before resource.
+	for _, p := range []struct {
+		prefix string
+		to     func(string) string
+	}{
+		{"resource.attributes.", ResourceAttributeField},
+		{"resources.", ResourceAttributeField},
+		{"resource.", ResourceAttributeField},
+		{"attributes.", SpanAttributeField},
+		{"events.", EventAttributeField},
```
faiz9068, line 326: **Minor: the singular `event.` sibling is still uncovered.** `resource.` and `resources.` are both handled, but only the plural `events.` is here — so `event.exception.type` falls through untouched, and `validateFieldSyntax` doesn't reject it either (it only catches `ResourceAttributes.`, `SpanAttributes.`, double-quoted brackets, and the flat `resource_`/`event_` prefixes).

Net effect is the same opaque `parse error: invalid JSON pipeline` this PR is closing out, for a form that's an easy thing for a model to emit. One more line in the table: `{"event.", EventAttributeField}`.

(Bare refs with no prefix at all — `db.system`, `deployment.environment` — also still reach the backend. Those genuinely shouldn't be guessed at, since the prefix is the only thing that says which column they belong to, so a descriptive `validateFieldSyntax` error is the better instrument there. Separate ticket.)

internal/telemetry/traces/query_sanitize.go
```diff
@@ -37,10 +37,10 @@ var traceFilterLogicalOperators = map[string]struct{}{
 	"$not": {},
 }
 
-// sanitizeTraceJSONQuery validates a tracejson_query pipeline before forwarding
+// SanitizeTraceJSONQuery validates a tracejson_query pipeline before forwarding
```
faiz9068, line 40: **Minor: the in-place-mutation contract got lost in the export.** This comment says the function "validates" and "returns descriptive errors", but it also mutates the caller's maps in place — `rewriteBrokenExistenceOperators`, `rewriteLegacyDotNotationFields`, and the top-level `$and` wrapping all write through the passed-in slice.

The deleted `SanitizeTraceJSONPipeline` documented exactly that ("Normalizations are applied in place") and that sentence didn't survive the consolidation. It matters more now, not less: the function is exported and `internal/apm` depends on the mutation for correctness — `buildDatabaseSlowQueryTracePipeline` returns the same slice it passed in. One sentence restores it.

internal/telemetry/traces/query_sanitize.go
```diff
@@ -56,6 +56,7 @@ func sanitizeTraceJSONQuery(stages []map[string]interface{}) error {
 				// ClickHouse). Rewrite both to the working idiom before
 				// validation ever sees them. Recurses through $and/$or/$not.
 				rewriteBrokenExistenceOperators(query)
+				rewriteLegacyDotNotationFields(query)
```
faiz9068, line 60: Ordering here is right and worth stating explicitly: `rewriteBrokenExistenceOperators` runs first, so the `$neq`s it synthesizes still get their fields normalized by this new pass. Swapping the two lines would leave a silent hole.

**Minor / follow-up:** this rewrite is gated on `stageType == "filter" || "where"` (line 52), so dot-notation map refs elsewhere in a pipeline still 400 — e.g. `{"type": "aggregate", "groupby": {"attributes.db.system": "db"}}`. Pre-existing scope limit rather than something this PR introduced, but the new rewrite makes the inconsistency more visible: the same field ref is auto-corrected in a filter stage and rejected in an aggregate stage.

internal/telemetry/traces/field_mapping.go
```diff
@@ -89,6 +91,21 @@ func enrichAttribute(raw string) TraceAttribute {
...
+	return enrichAttribute("resource_" + strings.TrimPrefix(key, "resource_")).FilterField
+}
+
+// EventAttributeField returns the tracejson filter field for a span event attribute key.
+func EventAttributeField(key string) string {
+	return enrichAttribute("event_" + strings.TrimPrefix(key, "event_")).FilterField
```
faiz9068, line 106: Both guards address the previous round's note, and routing through `enrichAttribute` keeps paying off — `ResourceAttributeField("service.name")` still correctly yields `ServiceName` rather than `resources['service.name']`, which is what makes the sanitizer handle `resource.attributes.service.name` right for free. Tests at `field_mapping_test.go:58-81` cover the prefixed and `service.name` cases.

internal/telemetry/traces/query_sanitize_test.go
```diff
@@ -804,3 +804,97 @@ func TestSanitizeTraceJSONQuery_RewritesBrokenExistenceOperators(t *testing.T) {
...
+			"ServiceName",
+			"Duration",
+		} {
+			if got := normalizeTraceFilterField(field); got != field {
+				t.Errorf("normalizeTraceFilterField(%q) = %q, want unchanged", field, got)
+			}
+		}
+	})
+}
+
```
faiz9068, line 905: **Nit:** trailing blank line — `gofmt -l internal/telemetry/traces/` flags this file. Nothing catches it in CI (`.github/workflows/test.yaml` runs only `go test -v -race ./...`), so it's cosmetic, but it'll show up as noise in the next person's diff.

While here — a note on what these new subtests do and don't pin: `TestGetDatabaseSlowQueriesHandler_TracePipelineUsesBracketNotation` would still pass if the `SpanAttributeField`/`ResourceAttributeField` change alone were reverted, because this sanitizer rewrites the dotted form back before dispatch. That's defence-in-depth working as designed, not a vacuous test — the builder-level test and these sanitizer subtests each fail independently if their own layer regresses. Just worth knowing the handler test isn't the thing guarding the builder.

faiz9068 (commented):
Re-reviewed `055f351..6923d6c`. **All three should-fix items from the last round are genuinely resolved**, and I verified the sanitizer by executing it rather than reading it — a throwaway probe ran `normalizeTraceFilterField` over ~40 inputs.

| Previous finding | Status |
|---|---|
| Sanitizer covered only 2 of 3 map columns | **Fixed** — `events.exception.type` → `events['exception.type']` (correct column, not `attributes`), both `resources.`/`resource.` variants → `resources['deployment.environment']`, `resource.attributes.service.name` → `ServiceName`, all 11 top-level fields unchanged, and `ResourceAttributes.`/`resource_` still reach `validateFieldSyntax`'s corrective errors instead of being silently swallowed |
| Handler test could pass vacuously | **Fixed** — mutex-guarded `reached` flag asserted on the test goroutine after the call; `-race` clean |
| `t.Fatalf` from the server goroutine | **Fixed** — all closure calls are `t.Errorf` + `return`; remaining `t.Fatal*` are all on the test goroutine |
| Dead conversion loop in `SanitizeTraceJSONPipeline` | **Fixed, better than suggested** — the wrapper was deleted outright and `sanitizeTraceJSONQuery` exported instead; `grep -rn SanitizeTraceJSONPipeline --include='*.go'` returns nothing |
| `ResourceAttributeField` unprefixed-key guard | **Fixed** — `TrimPrefix` on both `resource_` and `event_`, with tests |
| Missing CHANGELOG entry | **Fixed, and accurate** — the "`db_system`, `host` and `env` returned a 400" claim checks out against `9156bfa`, where the no-`db_system` `$neq` branch really was already bracketed |

The `refactor: move slow-query trace pipeline builder into helpers section` commit is provably behavior-preserving: a sorted-line diff of `databases.go` between `cbc0add` and `6923d6c` is empty, so it's a pure reorder with no token changed.

Remaining notes are all Minor and inline. One that has no diff line to anchor to: **`aggregate` stage `groupby` refs still bypass the rewrite** — worth a follow-up rather than a change here, noted inline at the call site.

Verification: `go build ./...` and `go vet ./...` clean; `go test -race ./internal/apm/ ./internal/telemetry/traces/` both ok with no race warnings; full `go test ./...` shows only the pre-existing `internal/dashboards` 403s from an unset `LAST9_DELETE_TOKEN`.

Nothing here blocks the fix. Leaving it as a comment rather than an approval.

internal/telemetry/traces/query_sanitize.go
faiz9068: Fixed in 394421d. Reproduced your five cases exactly before changing anything — a throwaway probe over `normalizeTraceFilterField` confirmed `"events."` → `attributes['event_']` and the rest.

Took the first option: skip the rewrite when the stripped remainder is empty, returning the field untouched. Chose it over relaxing the `enrichAttribute` length guards to `>=` because that would only move the garbage to the right column — `enrichAttribute("event_")` would yield `events['']` — while widening blast radius, since `enrichAttribute` also maps real API tag names for the discovery tools.

One correction to the suggestion: pass-through does not reach a `validateFieldSyntax` error. `"events."` matches none of its cases (`ResourceAttributes.`, `SpanAttributes.`, double-quoted brackets, flat `resource_`/`event_`), so nonsense input still gets the backend 400. What the fix buys is narrower than "useful sanitizer error": it guarantees the rewrite never picks the *wrong* map column, which was your actual concern. Adding a descriptive error for bare and prefix-only refs belongs with the no-prefix case you split out below.

New subtest pins all six prefix-only forms as unchanged.

internal/telemetry/traces/query_sanitize.go
faiz9068: Added in 394421d — `{"event.", EventAttributeField}`, with a subtest.

Worth noting I had this line in an earlier draft and deliberately removed it before the first push, on the grounds that you had enumerated only three escaping forms and `event.` was speculative surface. That was the wrong call: it is exactly as plausible a model emission as the `resource.` singular I did keep, and the asymmetry between the two was arbitrary.

Agreed on leaving bare refs (`db.system`) alone — the prefix is the only thing identifying the column, so guessing would be worse than the current 400. That wants a `validateFieldSyntax` error rather than a rewrite; will fold it into the same follow-up as the prefix-only error text.

internal/telemetry/traces/query_sanitize.go
faiz9068: Fixed in 394421d — one sentence restored: "Normalizations are applied in place — callers forward the slice they passed in."

Confirmed the dependency is load-bearing rather than incidental: `buildDatabaseSlowQueryTracePipeline` returns the same slice it handed to the sanitizer, so if the mutation stopped happening the builder would silently emit unsanitized dot notation and reintroduce the exact 400 this PR closes.

internal/telemetry/traces/query_sanitize.go
faiz9068: Noted, no change here — agreed it is a pre-existing scope limit rather than something this PR introduced.

Confirmed the gate: the rewrite sits inside `if stageType == "filter" || stageType == "where"`, so `groupby` keys, `transform` field refs and aggregate function args are all untouched. Folding it into one follow-up ticket with the two `validateFieldSyntax` gaps from the threads above — the common shape is that field refs outside filter stages get no normalization and no corrective error.

internal/telemetry/traces/query_sanitize_test.go
faiz9068: Fixed — `gofmt -w`, and `gofmt -l internal/telemetry/traces/` is now clean. (`internal/apm/deviations_types.go` is also unformatted, but that predates this branch and I left it alone.)

Agreed on the layering, and it is worth having stated explicitly: the handler test would survive reverting the builder change alone, because this sanitizer rewrites the dotted form before dispatch. `TestBuildDatabaseSlowQueryTracePipeline_UsesBracketNotation` is what actually guards the builder — it asserts on the pipeline pre-dispatch, so it fails on a builder regression regardless of the sanitizer.

faiz9068 (approved):

PR #204, opened 2026-08-06
Add guarded Alert Hygiene Pulse tools

prathamesh-sonpatki:
## Summary

- add primitive organization-scoped Pulse subscription, run, report, finding, evidence, and disposition tools
- forward canonical API JSON, pagination, truncation, status, version, audit, and HTTP conflict semantics without MCP-side state
- register eight read tools under `pulse_read` and five write tools under `pulse_manage`

## Safety

- Pulse writes remain absent for empty/default and `all` toolset selections
- writes require an explicit `pulse_manage` grant plus `confirmed: true`
- configuration and enablement are separate; no run-now, investigation launch, recommendation application, or alert mutation tool exists

## Verification

- full `go test -race ./...` passed
- full `go vet ./...` passed
- tool dumps prove read/manage isolation and fail-closed default-all behavior
- shared cross-surface `v1` fixture is checksum-pinned

## Dependency

- requires the canonical Pulse APIs from [last9/last9#11362](https://github.com/last9/last9/pull/11362)


internal/pulse/subscriptions.go
```diff
@@ -0,0 +1,141 @@
...
+
+func validateUpdate(args UpdateSubscriptionArgs) (string, error) {
+	if err := validateSubscriptionMutation(args.SubscriptionInput, args.Confirmed); err != nil {
+		return "", err
+	}
+	return escapedID(args.SubscriptionID)
+}
+
+func validateEnablement(args SetSubscriptionEnabledArgs) (string, error) {
+	if !args.Confirmed {
```
kneurgao, line 130: **[P1]** Confirmed-gate untested for update/enable/disable/disposition writes

`NewUpdateSubscriptionHandler`, `NewEnableSubscriptionHandler`, `NewDisableSubscriptionHandler`, and `NewWriteDispositionHandler` are never exercised with `Confirmed: false` in any test -- only `create_pulse_subscription`'s gate is asserted (`TestCreateSubscriptionRequiresConfirmation`). This is the PR's core safety mechanism; a future refactor could silently drop the check on one of these paths without any test catching it.

Suggested: add tests mirroring `TestCreateSubscriptionRequiresConfirmation` for the other three handlers (`Confirmed:false` -> error, 0 upstream requests), plus a `Confirmed:false` case for `write_pulse_disposition`.

internal/pulse/reports.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func NewGetReportHandler(httpClient *http.Client, config models.Config) func(context.Context, *mcp.CallToolRequest, GetReportArgs) (*mcp.CallToolResult, any, error) {
+	api := newClient(httpClient, config)
+	return func(ctx context.Context, _ *mcp.CallToolRequest, args GetReportArgs) (*mcp.CallToolResult, any, error) {
+		return api.getRunResource(ctx, args.RunID, "/report")
+	}
+}
+
+func NewListFindingsHandler(httpClient *http.Client, config models.Config) func(context.Context, *mcp.CallToolRequest, RunPageArgs) (*mcp.CallToolResult, any, error) {
```
kneurgao, line 63: **[P2]** `list_pulse_findings` and `get_pulse_finding` have zero test coverage

Neither `NewListFindingsHandler` nor `NewGetFindingHandler` is called by any test in `pulse_test.go`, `contract_test.go`, or `tools_test.go`, even though the checksum-pinned fixture already contains matching `findings_page`/`finding_detail` sections that `contractFixture` never decodes.

Suggested: add handler tests asserting path/cursor/limit forwarding, and wire the fixture's unused sections into `contractFixture` for a roundtrip check.

pulse_tools.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, register := range registrations {
+		if err := register(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func pulseReadTool(name string, action string, description string) *mcp.Tool {
+	return &mcp.Tool{Name: name, Description: fmt.Sprintf("%s.\n\n%s", action, description), Annotations: &mcp.ToolAnnotations{ReadOnlyHint: true}}
```
kneurgao: **[P2]** Tool description action labels are hardcoded Go strings, not markdown

`pulseReadTool`/`pulseWriteTool` take a hardcoded `action` string (e.g. "List subscriptions") and `fmt.Sprintf` it together with the go:embed'd markdown body. AGENTS.md's tool-description rule ("Never define description text as Go string constants in handler packages") is written without a short-label exception, so part of every served description here originates outside `internal/prompts/descriptions/`.

Suggested: fold the action line into each tool's own `.md` file (or a small per-tool addendum) instead of concatenating a Go literal at registration time.

internal/pulse/client.go
```diff
@@ -0,0 +1,163 @@
...
+		message = http.StatusText(status)
+	}
+	return &APIError{StatusCode: status, Body: message}
+}
+
+func textResult(body []byte) *mcp.CallToolResult {
+	return &mcp.CallToolResult{Content: []mcp.Content{&mcp.TextContent{Text: string(body)}}}
+}
+
+func escapedID(value string) (string, error) {
```
kneurgao, line 141: **[P2]** `escapedID` does not reject bare "."/".." path segments

The check only rejects values containing `/`, `?`, or `#`; a bare `".."` has none of those, passes through unescaped by `url.PathEscape` (which doesn't escape dots), and lands directly in the request path. The existing traversal test (`TestRunIDCannotChangeTheOrganizationScopedPath`) only covers a slash-containing value, so this specific case is uncovered.

Suggested: `if strings.ContainsAny(value, "/?#") || value == "." || value == ".." { return "", fmt.Errorf("ID contains invalid path characters") }`, or reject via `path.Clean`.

internal/pulse/reports.go
```diff
@@ -0,0 +1,123 @@
...
+
+type RunPageArgs struct {
+	RunID  string `json:"run_id" jsonschema:"(Required) Pulse run ID."`
+	Limit  int    `json:"limit,omitempty" jsonschema:"Page size from 1 to 100; defaults to 25."`
+	Cursor string `json:"cursor,omitempty" jsonschema:"Opaque next_cursor returned by the prior page."`
+}
+
+type GetFindingArgs struct {
+	RunID        string `json:"run_id" jsonschema:"(Required) Pulse run ID."`
+	OccurrenceID string `json:"occurrence_id" jsonschema:"(Required) Finding occurrence ID returned by list_pulse_findings."`
```
kneurgao: **[P2]** `get_pulse_finding`'s `occurrence_id` doesn't match any `list_pulse_findings` key

The fixture's `findings_page` items expose the value this field wants under the key `id`, while a separate, more prominent `finding_id` key points at a different resource. The jsonschema description just says "returned by list_pulse_findings" without naming the literal key, so an LLM caller chaining these two tools will foreseeably copy `finding_id` instead and get a 404.

Suggested: `jsonschema:"(Required) The 'id' field (not 'finding_id') of a list_pulse_findings item."`

internal/pulse/subscriptions.go
```diff
@@ -0,0 +1,141 @@
...
+
+func NewUpdateSubscriptionHandler(httpClient *http.Client, config models.Config) func(context.Context, *mcp.CallToolRequest, UpdateSubscriptionArgs) (*mcp.CallToolResult, any, error) {
+	api := newClient(httpClient, config)
+	return func(ctx context.Context, _ *mcp.CallToolRequest, args UpdateSubscriptionArgs) (*mcp.CallToolResult, any, error) {
+		id, err := validateUpdate(args)
+		if err != nil {
+			return nil, nil, err
+		}
+		body, err := api.call(ctx, request{method: http.MethodPut, path: "/subscriptions/" + id, body: args.SubscriptionInput})
+		return handlerResult(body, err)
```
kneurgao, line 85: **[P2]** Full-replace subscription update has no conflict detection

`write_pulse_disposition` has explicit optimistic concurrency (`expected_version`, 409 on stale write, tested). `update_pulse_subscription` has no equivalent -- `SubscriptionInput` carries no version field and this sends a plain `PUT` with no conditional header, so two concurrent callers can silently clobber each other's subscription config.

pulse_tools.go
```diff
@@ -0,0 +1,101 @@
...
+
+func registerPulseTools(server *last9mcp.Last9MCPServer, config models.Config, client *http.Client) error {
+	for _, register := range []func() error{
+		func() error { return registerPulseSubscriptions(server, config, client) },
+		func() error { return registerPulseReports(server, config, client) },
+		func() error { return registerPulseDisposition(server, config, client) },
+	} {
+		if err := register(); err != nil {
+			return err
+		}
```
kneurgao: **[P2]** Pulse registration reinvents a closure-list wrapper 3x

`registerPulseSubscriptions`, `registerPulseReports`, and `registerPulseDisposition` each build a `[]func() error{...}` slice of single-line closures purely to hand to a bespoke `runRegistrations` loop -- every closure just forwards its arguments, and the same three-line scaffold repeats three times for 13 registrations, inconsistent with the flat `reg(registerIfAllowed(...))` pattern used everywhere else in `tools.go`.

Suggested: drop `runRegistrations`, call `registerIfAllowed` directly in a flat sequence (or reuse the `reg`/`regErr` accumulator tools.go already uses elsewhere).

kneurgao (commented):
7 new findings:
- P1 — Confirmed-gate untested for update/enable/disable/disposition writes (`internal/pulse/subscriptions.go:130`)
- P2 — Missing test coverage for list_pulse_findings/get_pulse_finding (`internal/pulse/reports.go:63`)
- P2 — Tool description action labels are hardcoded Go strings, not markdown (`pulse_tools.go:95`)
- P2 — escapedID doesn't reject bare "."/".." path segments (`internal/pulse/client.go:141`)
- P2 — get_pulse_finding's occurrence_id doesn't match list_pulse_findings' key (`internal/pulse/reports.go:34`)
- P2 — Full-replace subscription update has no conflict detection (`internal/pulse/subscriptions.go:85`)
- P2 — Pulse registration reinvents a closure-list wrapper 3x (`pulse_tools.go:24`)

internal/pulse/subscriptions.go
prathamesh-sonpatki: > **[P1]** Confirmation gates lack direct coverage.

Addressed in `21bcad2`: update, enable, disable, and disposition handlers reject `confirmed:false` and make zero upstream requests.

internal/pulse/reports.go
prathamesh-sonpatki: > **[P2]** Findings list/detail handlers lack coverage.

Addressed in `21bcad2`: exact path/cursor/limit tests and canonical fixture roundtrips cover both.

pulse_tools.go
prathamesh-sonpatki: > **[P2]** Action labels are hardcoded outside markdown.

Addressed in `21bcad2`: action copy is embedded in the three Pulse markdown descriptions and served directly.

internal/pulse/client.go
prathamesh-sonpatki: > **[P2]** Bare dot path segments are accepted.

Addressed in `21bcad2`: both `.` and `..` are rejected with regression tests.

internal/pulse/reports.go
prathamesh-sonpatki: > **[P2]** The finding occurrence argument is ambiguous.

Addressed in `21bcad2`: the schema explicitly requires the list item's `id`, not `finding_id`, while retaining `occurrence_id` to match the API route.

internal/pulse/subscriptions.go
prathamesh-sonpatki: > **[P2]** Subscription replacement lacks optimistic concurrency.

Not changed: the API provides no revision, ETag, `If-Match`, or `expected_version`; `config_version` is a schema version. This requires upstream conditional-write support before MCP can enforce it.

pulse_tools.go
prathamesh-sonpatki: > **[P2]** Pulse registration repeats a closure wrapper.

Addressed in `21bcad2`: one typed registrar preserves read/manage boundaries and first-error propagation.

pulse_tools.go
```diff
@@ -0,0 +1,68 @@
...
+
+func registerPulseTools(server *last9mcp.Last9MCPServer, config models.Config, client *http.Client) error {
+	registrar := &pulseRegistrar{server: server}
+	registerPulseSubscriptions(registrar, config, client)
+	registerPulseReports(registrar, config, client)
+	registerPulseDisposition(registrar, config, client)
+	return registrar.err
+}
+
+type pulseRegistrar struct {
```
kneurgao: **[P2]** Pulse registration reinvents `reg`/`regErr` instead of reusing it

`tools.go`'s `registerAllTools` already threads a single `reg(err)` closure over `regErr` through every other tool registration. This adds a second, parallel implementation of the same "accumulate first error, short-circuit the rest" idiom via `pulseRegistrar` + `registerPulseHandler`, only handing the aggregate to `reg` at the very end.

Suggested: thread the existing `reg func(error)` closure into `registerPulseTools` and call `reg(registerIfAllowed(...))` directly and flatly for each of the 13 pulse tools, matching every other tool in `tools.go`. Delete `pulseRegistrar`/`registerPulseHandler`.

internal/toolsets/manage_allowlist.go
```diff
@@ -0,0 +1,38 @@
+package toolsets
+
+const (
+	manageGrantMarker = "\x00pulse_manage_explicit"
+	allToolsMarker    = "\x00all_tools"
+)
+
+// AllowsManage requires both normal toolset membership and an explicit
+// pulse_manage token. Empty and default-all selections never grant writes.
+func AllowsManage(allowed Set, toolName string) bool {
```
kneurgao: **[P3]** `AllowsManage` is exported with zero production callers

Nothing in `pulse_tools.go`, `subscriptions.go`, or `dispositions.go` calls `AllowsManage` — the only caller in the whole diff is its own test. The real gating happens once, at registration time, via `ManageOnly`. A reader encountering `AllowsManage` in the package's public API will reasonably assume a per-call runtime check exists.

Suggested: either delete `AllowsManage` and its test (registration-time gating already covers this), or actually call it from the top of each write handler if a defense-in-depth check was intended, with a handler-level test proving it rejects when the grant is absent.

kneurgao (commented):
Nice cleanup — verified all 6 resolved threads against the new code rather than just the summary: the confirmed-gate tests, findings-handler coverage, markdown-only descriptions, escapedID dot-segment rejection, and the occurrence_id disambiguation all landed as described, each with a real regression test. The registration refactor also genuinely reduced the 3x closure-list duplication down to one helper.

3 new findings (1 older still open):
- P2 — Pulse registration reinvents `reg`/`regErr` instead of reusing it (`pulse_tools.go:23`)
- P3 — `AllowsManage` is exported with zero production callers (`internal/toolsets/manage_allowlist.go:10`)
- P3 — Shared per-tool-group markdown still gives 6 distinct tools one identical description (`internal/prompts/descriptions/pulse_subscriptions.md:3`)

internal/prompts/descriptions/pulse_subscriptions.md
```diff
@@ -0,0 +1,5 @@
+Manage Alert Hygiene Pulse subscription configuration through canonical organization-scoped APIs.
+
+Actions: list subscriptions (`list_pulse_subscriptions`), get a subscription (`get_pulse_subscription`), create a disabled subscription (`create_pulse_subscription`), update a subscription (`update_pulse_subscription`), enable it (`enable_pulse_subscription`), or disable it (`disable_pulse_subscription`).
```
kneurgao: **[P3]** Shared per-tool-group markdown still gives 6 distinct tools one identical description

This (and `pulse_reports.md`) genuinely moved out of hardcoded Go strings, resolving the round-1 finding as literally asked. But the same multi-action string is still reused byte-for-byte as the `Tool.Description` for all 6 tools in the group — e.g. `get_pulse_run`/`get_pulse_report` (single-object gets) ship `pulse_reports.md`'s "Follow opaque `next_cursor`..." guidance despite having no cursor param. Low practical risk (an LLM caller would just not find the field and self-correct), so advisory rather than blocking.

Suggested: split into one description file per tool (mirroring `get_dashboard_snapshot.md`), with genuinely shared guardrail language (confirmed:true, pulse_manage) factored into a short paragraph appended to each per-tool file.

pulse_tools.go
prathamesh-sonpatki: > Pulse registration reinvents the existing `reg`/`regErr` accumulator.

Addressed: Pulse registration now receives the existing `reg func(error)` and calls `registerIfAllowed` directly; the duplicate registrar types were removed.

internal/toolsets/manage_allowlist.go
prathamesh-sonpatki: > `AllowsManage` is exported with zero production callers.

Addressed: removed `AllowsManage` and updated coverage to exercise the production `ManageOnly` registration view. Fail-closed managed-tool gating remains covered.

internal/prompts/descriptions/pulse_subscriptions.md
prathamesh-sonpatki: > Shared per-tool-group markdown still gives distinct tools one identical description.

Addressed: all six subscription tools and all six report tools now use action-specific embedded markdown descriptions. Registration tests verify exact wiring and uniqueness.

internal/prompts/descriptions/list_pulse_subscriptions.md
```diff
@@ -0,0 +1,3 @@
+List Alert Hygiene Pulse subscriptions for the configured organization.
```
kneurgao, line 1: **[P3]** Kept the old group filename instead of the tool's own name

This file now backs exactly one tool (`list_pulse_subscriptions`) rather than all 6 subscription tools — the split itself is genuinely done. But unlike its 5 siblings (`create_pulse_subscription.md`, `update_pulse_subscription.md`, etc.), this one wasn't renamed to match its tool. Same pattern at `pulse_reports.md` (backs only `list_pulse_runs` now) and `pulse_dispositions.md` (backs only `write_pulse_disposition`).

Suggested: rename to `list_pulse_subscriptions.md` (and the other two analogously), updating the `go:embed` paths in `prompts.go`.

internal/prompts/descriptions/list_pulse_runs.md
```diff
@@ -0,0 +1,3 @@
+List Alert Hygiene Pulse runs for the configured organization.
```
kneurgao, line 1: **[P3]** Same as `pulse_subscriptions.md` — this file now backs only `list_pulse_runs` but wasn't renamed to `list_pulse_runs.md`.

internal/prompts/descriptions/write_pulse_disposition.md
```diff
@@ -0,0 +1,5 @@
+Write one optimistic-concurrency Alert Hygiene Pulse disposition: `tune`, `expected`, or `investigate`.
```
kneurgao, line 1: **[P3]** Same pattern — this file now backs only `write_pulse_disposition` but wasn't renamed to `write_pulse_disposition.md`.

kneurgao (commented):
All 3 round-2 findings independently verified as genuinely resolved, not reshuffled: `pulse_tools.go` registration now threads the same `reg`/`regErr` closure `tools.go` already uses (no `pulseRegistrar` left anywhere), `AllowsManage` is gone and replaced by `ManageOnly` with real production callers gating all 5 write tools, and the description split produced 13 distinct one-per-tool files with spot-checked-accurate content.

1 new finding (across 3 files; 1 older thread still open):
- P3 — 3 of the 13 renamed description files kept their old group filename instead of matching their own tool (`pulse_subscriptions.md`, `pulse_reports.md`, `pulse_dispositions.md`)

internal/prompts/descriptions/write_pulse_disposition.md
prathamesh-sonpatki: > This file now backs only `write_pulse_disposition` but wasn't renamed to `write_pulse_disposition.md`.

Addressed: renamed it to `write_pulse_disposition.md` and updated the embed path; content and exported Go API are unchanged.

internal/prompts/descriptions/list_pulse_subscriptions.md
prathamesh-sonpatki: > Kept the old group filename instead of the tool's own name.

Addressed: renamed the three remaining group files to `list_pulse_subscriptions.md`, `list_pulse_runs.md`, and `write_pulse_disposition.md`, and updated their embed paths. Exported Go variables remain unchanged to avoid unrelated API churn.

internal/prompts/descriptions/list_pulse_runs.md
prathamesh-sonpatki: > This file now backs only `list_pulse_runs` but wasn't renamed to `list_pulse_runs.md`.

Addressed: renamed it to `list_pulse_runs.md` and updated the embed path; content and exported Go API are unchanged.

PR #210, opened 2026-08-13
fix(apm): rank get_service_summary by interval counts

prathamesh-sonpatki:
## Summary

- Makes `get_service_summary` the only fleet ranking surface: ranked `(service, env)` rows with interval `request_count`, derived `throughput_rpm`, and separate `http_4xx_count` / `http_5xx_count` / `grpc_error_count`.
- Drops `ErrorRate` and the p95-of-counts queries. Sort/limit/rank run in process (no PromQL `topk`). Unknown `sort_by` (including `errors`) is rejected. Empty vectors stay `rows: []`; any of the four Prom queries failing fails the whole call.
- Closes the ENG-1713 ranking contract and ENG-1722 `ErrorRate` removal. Routing evals stay on ENG-1712.

Fixes ENG-1713. Folds in ENG-1722.

## Test plan

- [x] `go test ./internal/apm/ -count=1 -run 'TestNewServiceSummaryHandler'` (AE1–AE6, PromQL shape, fail-whole, empty rows, limit clamp)
- [x] `go test ./internal/utils/ -count=1 -run 'TestGetTimeRange'`
- [x] `go test . -count=1 -run 'TestDumpTools'` plus `go run . dump-tools` (schema has `sort_by`/`limit`, no `ErrorRate`) and `go run . dump-tools --toolsets=alerts` (tool absent)
- [x] `go test ./internal/prompts/ -count=1 -run 'TestAPMServiceDeviationsDescription'`
- [x] `go test ./internal/apm/ ./internal/prompts/ .`
- [ ] Hosted MCP replay of an anonymized “top 10 by 5xx” prompt after deploy (release check, not this PR)
- [ ] ENG-1712 evals / repeated-prompt routing — **waived here** (separate ticket)

## Post-Deploy Monitoring & Validation

- Watch MCP tool traces for `get_service_summary`: error rate should not rise from unknown `sort_by` (expected client errors) or Prom class failures (fail-whole is intentional).
- Confirm a fleet call returns `rows` (not a service-name map) with `sort_by=request_count` when omitted, and that `sort_by=http_5xx_count` changes order.
- Failure signal: clients still unmarshaling PascalCase `ErrorRate` / map keyed by service. Rollback: revert this PR.
- Validation window: first hosted MCP session after deploy; owner: MCP.

## Known Residuals

- `p95_latency` omitted until a query matches Service Details performance-history charts.
- README “What It Can Do” one-liner for this tool still mentions error rate / p95 at a high level; only the env default was corrected per plan U4.
- Full `ce-code-review` persona roster was substituted with a single in-harness reviewer (lost independent multi-persona coverage).


internal/apm/service_summary.go
```diff
@@ -0,0 +1,339 @@
...
+	if strings.HasPrefix(scope, "^") && strings.HasSuffix(scope, "$") && len(scope) >= 2 {
+		inner = strings.TrimSuffix(strings.TrimPrefix(scope, "^"), "$")
+	}
+	if inner == "" || strings.ContainsAny(inner, serviceSummaryRegexMeta) {
+		return ""
+	}
+	return inner
+}
+
+func intervalMinutes(startUnix, endUnix int64) int {
```
faiz9068, line 268: **The reported interval can differ from the interval actually summed.**

`intervalMinutes` rounds *up*, and that rounded value becomes both the PromQL `[Nm]` range and the `throughput_rpm` divisor — which is self-consistent, and correct for making sure the range covers the whole window. But `start_time` / `end_time` in the response envelope are the *requested* bounds, so the two can disagree. Two reachable cases:

- `GetTimeRange` (`internal/utils/utils.go:127`) guards only `startTime.After(endTime)`, so `start == end` is accepted. That yields `windowMin = 1`: the tool sums a full minute of data and reports a zero-length interval.
- A 90-second window yields `[2m]` — 120s of data reported as a 90s window.

For a change whose whole point is that this tool stops misreporting what it measured, this is the one place where it still does. Two ways to close it, not mutually exclusive: emit `window_minutes` in the response (cheap, and it makes the `throughput_rpm` denominator auditable), and/or snap `start_time` to `end_time - windowMin` so the stamped interval is the one actually queried.

There's also no test for the rounding path — the only rpm assertion uses an exact 20m window, so neither the sub-minute nor the non-integer case is covered.

internal/apm/service_summary.go
```diff
@@ -0,0 +1,339 @@
...
+	Coverage         string              `json:"coverage"`
+	Hint             string              `json:"hint,omitempty"`
+}
+
+type ServiceSummaryArgs struct {
+	StartTimeISO    string  `json:"start_time_iso,omitempty" jsonschema:"Start of the interval in RFC3339/ISO8601 (e.g. 2024-06-01T12:00:00Z). When both start and end are set they beat lookback."`
+	EndTimeISO      string  `json:"end_time_iso,omitempty" jsonschema:"End of the interval in RFC3339/ISO8601 (e.g. 2024-06-01T13:00:00Z). When both start and end are set they beat lookback. A single bound fills the other with lookback_minutes."`
+	LookbackMinutes float64 `json:"lookback_minutes,omitempty" jsonschema:"Number of minutes to look back from now (default: 60, minimum: 1). Use for relative windows like last 30 minutes."`
+	Env             string  `json:"env,omitempty" jsonschema:"Environment PromQL regex (default: .*). Exact one-env match needs anchors (e.g. ^prod$)."`
+	SortBy          string  `json:"sort_by,omitempty" jsonschema:"Sort key. Allowed: request_count (default), throughput_rpm, http_4xx_count, http_5xx_count, grpc_error_count. Unknown values including errors, error_rate, and 5xx are rejected."`
```
faiz9068: **`sort_by`'s five legal values live only in this prose, not in the served schema.**

Struct-tag generation emits a bare `string` for `sort_by` (and no `minimum`/`maximum` for `limit` below), so a client reading `tools/list` gets no enum and can't constrain the call before making it.

The handler rejects unknown values correctly — but a schema `enum` *prevents* the bad call rather than failing it after a round trip, which matters more for a tool whose caller is a model. The machinery already exists: `stringEnum()` at `internal/telemetry/traces/input_schema.go:3`, used by hand-written schemas attached at four `tools.go` call sites.

Worth noting `dump_test.go:79` only asserts these two properties *exist*, so it would pass unchanged if they stayed unconstrained.

internal/apm/service_summary.go
```diff
@@ -0,0 +1,339 @@
...
+	envMatcher := fmt.Sprintf(`env=~"%s"`, serviceSummaryFingerprintEnv)
+	parts := make([]string, 0, len(serviceSummaryQueried))
+	for _, class := range serviceSummaryQueried {
+		parts = append(parts, class.query(envMatcher, windowMin))
+	}
+	sum := sha256.Sum256([]byte(strings.Join(parts, "\n")))
+	return hex.EncodeToString(sum[:])
+}
+
+func serviceSummaryCatalogEnv(scope string) string {
```
faiz9068: **Deep-link env sanitation: right idea, two issues.**

**(a) Wrong layer.** No other `deeplink` caller sanitises its input. The two sibling `BuildAPMServiceLink` calls at `apm.go:523` and `apm.go:1065` pass raw `env` straight through, so `.*` currently ends up in the UI filter from those tools. Moved into `deeplink.BuildAPMServiceLink` itself, this logic would fix both for free; left local, the next tool that builds an APM link either re-derives it or ships the same bug.

**(b) Unanchored env still produces a mismatched link.** For `env="prod"` this returns `"prod"` and stamps an exact catalog filter, but the query ran `env=~"prod"` and may have matched `production` too — precisely the trap the tool description warns callers about. So `rows` and the dashboard link can describe different sets. Emitting a literal only for the anchored `^name$` form (falling back to no filter otherwise) keeps them consistent.

faiz9068 (commented):
Reviewed the cumulative diff (both commits), traced the shared helpers (`escapePromQLLabel`, `identityLess`, `GetTimeRange`, `MakePromInstantAPIQuery`), and ran `go build` / `go vet` / `go test ./internal/apm/ ./internal/prompts/ .` — all green.

Most of this holds up under scrutiny. No correctness bug in the ranked output. Four things I'd fix before merge; three are inline below.

### What checks out (verified, not assumed)

- **`sum_over_time` is the right aggregation here.** Checked against a live deployment: `trace_endpoint_count` behaves as a per-interval delta at 60s resolution, so `sum_over_time(...[1m])` returns exactly one sample's worth — no double-count. It also matches the existing convention at `apm.go:442,554,590`.
- **The gRPC matcher is safe.** `grpc_status_code!=""` correctly prevents the classic "negative matcher also matches series missing the label" inflation; confirmed against live data.
- **`escapePromQLLabel` closes the injection surface** — it escapes `\` before `"`, so a crafted `env` stays inside the string literal, and because PromQL unescapes `\\d` back to `\d`, legitimate regex escapes still work. The documented regex semantics survive.
- **`truncated` against the clamped limit is the honest signal** — a caller asking for 500 gets `limit: 100, truncated: true`, which correctly says "there is more and you cannot have it."
- **The tests assert behaviour, not the implementation.** Several would actually catch a wrong handler: the default-sort case asserts `request_count` ordering *despite* a large 5xx on another row; the 4xx/5xx case asserts the class queries don't cross-contaminate; `FailsWholeCallOnClassError` asserts the gRPC query never runs after a 5xx failure.

Dropping `ErrorRate` (a p95 of a count wearing a rate's name) and grouping by `(service_name, env)` are both clear wins — the old map collapsed multi-env services into one entry.

---

### 1. Missing `CHANGELOG.md` entry

Commenting here rather than inline since the file isn't in the diff. `## [Unreleased]` is currently empty, but #202, #201, #197, #191, #186 and #183 each added an entry.

This is the PR that most needs one: the response shape goes from `map[string]ServiceSummary` with PascalCase keys to `{rows: [...]}` with snake_case. The PR description already names old-shape unmarshaling as the rollback signal — that's exactly what a changelog entry communicates to anyone consuming the tool. Suggest a `### Changed` bullet at the top of its section, explicitly calling out the break.

---

### Nits (non-blocking)

- **Undocumented response fields.** `coverage`, `sort_key_unit`, `query_fingerprint`, `env_scope` and `row_count` aren't mentioned in the tool description — only `rows`, `truncated` and `limit` are. An agent sees fields it has no guidance on interpreting.
- **`sort_by=throughput_rpm` is ranking-identical to `request_count`.** rpm is the count divided by a constant, so the two produce the same order; the only observable difference is `sort_key_unit`. The description presents them as distinct choices, so agents will spend a decision on it for no ordering benefit. Worth either saying so explicitly or dropping the key.

internal/apm/service_summary.go
prathamesh-sonpatki: Fixed in f3c2e83: response now includes `window_minutes`, and `start_time` is snapped to `end_time - window_minutes` so the stamped interval matches the PromQL `[Nm]` range / rpm divisor. Added rounding-path unit tests (0-length, 90s → 2m).

internal/apm/service_summary.go
prathamesh-sonpatki: Fixed in f3c2e83: `GetServiceSummaryInputSchema()` now serves `sort_by` as an enum of the five keys and `limit` with min 0 / max 100; attached at registration in `tools.go`. `dump_test.go` asserts the enum length and bounds.

internal/apm/service_summary.go
prathamesh-sonpatki: Fixed in f3c2e83:
(a) Sanitation lives in `deeplink.APMCatalogEnv` / `BuildAPMServiceLink` — `.*` and regex patterns are skipped for every APM caller; plain exact tokens still work for sibling tools.
(b) `get_service_summary` only passes anchored `^name$` into the link builder, so unanchored `prod` no longer stamps an exact catalog filter that disagrees with the PromQL match set.

prathamesh-sonpatki:
Addressed the review in f3c2e83:

1. **CHANGELOG** — `### Changed` breaking note for the ranked `{rows}` envelope vs the old PascalCase map.
2. **Honest interval** — `window_minutes` + snapped `start_time`; rounding tests added.
3. **Schema** — hand-written `sort_by` enum + `limit` bounds on `tools/list`.
4. **Deeplink env** — sanitation centralized in `BuildAPMServiceLink`; summary only links anchored `^name$`.

Nits also landed in the tool description: envelope fields documented, and `throughput_rpm` called out as ranking-identical to `request_count`.

internal/deeplink/deeplink.go
```diff
@@ -114,6 +115,32 @@ func (b *Builder) BuildDropRulesLink() string {
...
+		return ""
+	}
+	if len(env) >= 3 && strings.HasPrefix(env, "^") && strings.HasSuffix(env, "$") {
+		inner := env[1 : len(env)-1]
+		if inner == "" || strings.ContainsAny(inner, apmCatalogEnvRegexMeta) {
+			return ""
+		}
+		return inner
+	}
+	if strings.ContainsAny(env, apmCatalogEnvRegexMeta) {
```
faiz9068: **This rejection is right for `get_service_summary` but is a small regression for the other four callers.**

Moving the sanitiser here was the correct call — but `get_service_summary` is the only caller whose `env` is a *regex*. The other four pass an exact value: `apm.go:479` interpolates it as `env='%s'` (exact match, not `=~`), and `deviations.go:199` passes `args.Env` straight through. For those, a metacharacter in the string isn't a pattern — it's just part of the environment's name.

So any env name containing a `.` now silently loses its catalog filter, where before this commit it got a working one. Verified by execution against `f3c2e83`:

```
env="prod.us-east"   APMCatalogEnv=""               filterPresent=false
env="k8s.prod"       APMCatalogEnv=""               filterPresent=false
env="team(a)"        APMCatalogEnv=""               filterPresent=false
env="prod"           APMCatalogEnv="prod"           filterPresent=true
env="prod-us-east-1" APMCatalogEnv="prod-us-east-1" filterPresent=true
```

Dotted environment names are common enough (`k8s.prod`, `prod.us-east`) that this will show up as "the deep link stopped filtering" on `get_service_performance_details`, `get_service_operations_summary`, `get_service_dependency_graph` and `get_apm_service_deviations` — none of which this PR otherwise touches.

The cleanest split is to let the caller declare which kind of input it has: keep this regex-rejecting path for the regex caller, and give the exact-value callers a path that only screens out `""`/`.*` and passes the rest through as a literal. Concretely, either export a second helper (`APMCatalogEnvExact`) for them, or have `BuildAPMServiceLink` take an already-decided literal and let each caller choose its own sanitiser — which is closer to what `serviceSummaryDeeplinkEnv` already does on the summary side.

Non-blocking — it degrades a link, not a result.

faiz9068 (commented):
Re-reviewed the delta `36e0250..f3c2e83` (not the whole PR again). All four items from my last review are addressed, and I verified each by execution rather than by reading the commit message. Head-commit CI is green on all six checks; `go build` / `go vet` / `go test ./internal/apm/ ./internal/deeplink/ ./internal/prompts/ .` pass at `f3c2e83`.

### Resolved — verified

1. **CHANGELOG** — `### Changed` bullet under `[Unreleased]`, correctly marked breaking and naming both the removed fields and the new params. Top of its section, per convention. ✅

2. **Window honesty** — this is a better fix than what I suggested. `window_minutes` is now in the envelope *and* `start_time` is stamped from `queriedStart = end - windowMin*60` (`service_summary.go:119`), so the reported interval is the one actually summed rather than the requested bounds. `TestIntervalMinutesRoundsUp` covers the exact-, zero-, 90s- and 1s-cases, and `TestNewServiceSummaryHandler_StampsQueriedWindow` asserts `window_minutes=2`, the stamped start, `throughput_rpm = 30/2 = 15`, **and** that the query carries `[2m]`. ✅

   Worth noting that last assertion closes a hole I hadn't reported: before this commit no test checked the range selector, so a handler that hard-coded `[30m]` would have passed the whole suite. It now wouldn't.

   One deliberate residual: `start == end` still isn't rejected — it becomes a 1-minute query. But it's now *disclosed* as a 1-minute window instead of being reported as zero-length, which was the actual complaint. Fine as-is.

3. **`sort_by` enum** — `GetServiceSummaryInputSchema()` with a real `enum` derived from `serviceSummarySortSpecs` (so it can't drift from the handler) plus `minimum`/`maximum` on `limit`, wired at `tools.go:70`. `dump_test.go` asserts 5 enum values and the 0/100 bounds, so a future divergence fails CI. ✅

4. **Deep-link env** — both parts fixed, and the layering fix went the whole way: the logic now lives in `deeplink.APMCatalogEnv`, so all five `BuildAPMServiceLink` callers get it rather than just this one. `serviceSummaryDeeplinkEnv` correctly requires anchors, so unanchored `prod` no longer stamps an exact filter that could disagree with rows. ✅

Both nits are handled too — `throughput_rpm`'s ranking-equivalence is now stated in the schema, the arg tag and the description, and the response envelope has a documented field list.

### One new issue introduced by fix 4

Inline below. Short version: sharing the sanitiser was right, but the four *other* callers pass an **exact** env, not a regex, so the regex rejection now silently drops the filter for legitimate env names containing a `.`. Non-blocking, but it's a small behaviour regression for tools this PR otherwise doesn't touch.

### Still open from my earlier review (both minor, neither blocking)

- **All four class-query failures share one opaque error** — `service_summary.go:298`: `"failed to get service summary: %s"`. With four near-identical queries in a loop, a 400 on the gRPC matcher and a 500 on the base query are indistinguishable to the agent and to whoever debugs it. Threading `class.key` into `fetchPromInstantSeries` is a one-line fix.
- **`truncated: true` doesn't say how much was dropped** — `row_count` is post-truncation, so the description's "raise limit (max 100) and recall" gives the agent no basis for picking a new limit. Returning the pre-truncation total would make that instruction actionable.

Also still there, trivially: the `env == ".*"` branch in `serviceSummaryEnvMatcher` (`:233`) is dead — it produces byte-identical output to the general path, since `escapePromQLLabel(".*") == ".*"`.

Nothing here blocks merge from my side.

internal/deeplink/deeplink.go
prathamesh-sonpatki: Fixed in a056312:
- `BuildAPMServiceLink` now uses `APMCatalogEnvExact` (only screens `""` / `.*`), so dotted exact names like `k8s.prod` keep their catalog filter for the other APM tools.
- `get_service_summary` pre-sanitizes with `APMCatalogEnvFromRegex` (anchored `^name$` only).

Also landed the two minor notes: Prom errors include the failing class key, and the envelope adds `matched_count` (pre-truncation total) so truncated recalls can pick a limit. Removed the dead `env == ".*"` branch in the matcher.

prathamesh-sonpatki:
Second-round follow-ups in a056312:

- **Deeplink:** `APMCatalogEnvExact` for exact-env callers; `APMCatalogEnvFromRegex` for summary — dotted names like `k8s.prod` keep filters again.
- **Errors:** Prom failures name the class key (`http_5xx_count`, etc.).
- **Truncation:** `matched_count` is the pre-limit total; description updated.
- **Cleanup:** dead `.*` matcher branch removed.

internal/deeplink/deeplink.go
```diff
@@ -114,8 +115,46 @@ func (b *Builder) BuildDropRulesLink() string {
...
+	inner := env[1 : len(env)-1]
+	if inner == "" || strings.ContainsAny(inner, apmCatalogEnvRegexMeta) {
+		return ""
+	}
+	return inner
+}
+
+// APMCatalogEnv is an alias of APMCatalogEnvFromRegex kept for older call sites.
+// Prefer APMCatalogEnvExact or APMCatalogEnvFromRegex explicitly.
+func APMCatalogEnv(env string) string {
```
faiz9068: Dead code — zero callers in the tree, and the "older call sites" it's kept for don't exist: `APMCatalogEnv` was added in `f3c2e83` and its only caller was rewired to `APMCatalogEnvExact` in this commit. Worth deleting rather than shipping an exported alias that invites the ambiguity the split just removed.

faiz9068 (commented):
Re-reviewed `f3c2e83..a056312`. All three remaining items are fixed and verified: per-class error keys (asserted in `FailsWholeCallOnClassError`), `matched_count` as the pre-truncation total, and the `Exact`/`FromRegex` split that restores the filter for dotted env names — `TestAPMCatalogEnvExact` now pins `k8s.prod` / `prod.us-east` / `team(a)`. I also confirmed by execution that dropping the `.*` branch in `serviceSummaryEnvMatcher` is behaviour-preserving on every path.

CI green at `a056312`; `go build` / `go vet` / `go test ./...` pass locally.

One dead function left, inline. Nothing blocking.

internal/deeplink/deeplink.go
prathamesh-sonpatki: Removed in 6baeee9 — `APMCatalogEnv` is gone; only `APMCatalogEnvExact` and `APMCatalogEnvFromRegex` remain.

internal/apm/service_summary_input_schema.go
```diff
@@ -0,0 +1,45 @@
...
+			},
+			"sort_by": map[string]interface{}{
+				"type":        "string",
+				"enum":        sortKeys,
+				"description": "Sort key. Allowed: request_count (default), throughput_rpm, http_4xx_count, http_5xx_count, grpc_error_count. throughput_rpm ranks identically to request_count (same order; only sort_key_unit differs). Unknown values including errors, error_rate, and 5xx are rejected.",
+			},
+			"limit": map[string]interface{}{
+				"type":        "integer",
+				"minimum":     float64(0),
+				"maximum":     float64(serviceSummaryMaxLimit),
```
faiz9068: **The schema and its own description contradict each other, and the schema wins.** `maximum: 100` here versus "Values above 100 clamp to 100" on the next line (and in `get_service_summary.md`).

Verified through the real in-memory SDK session, the same path `schema_validation_test.go` uses:

```
limit=150  IsError=true   validating /properties/limit: maximum: 150 is greater than 100
limit=-1   IsError=true   validating /properties/limit: minimum: -1 is less than 0
limit=100  IsError=false  (reaches the handler)
```

So the clamp branch in `resolveServiceSummaryLimit` is dead for every real MCP client, and `TestNewServiceSummaryHandler_LimitClampAndRanks`'s `Limit: 101` case passes only because it calls the handler directly and bypasses the served schema — the test asserts behaviour no client can reach. Same story one line up: "Other values below 1 are an error" is true of the handler, but `minimum: 0` rejects them first.

A model that follows the description and sends `limit: 150` gets a hard validation error instead of 100 rows. Fix either direction — drop `maximum`/`minimum` and keep the clamp (friendlier for an LLM caller, and the description then stays true), or keep the bounds and change both description strings to "must be between 1 and 100". 

My fault for asking for the bounds in the first round without re-reading the prose they'd contradict.

internal/apm/service_summary.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	return nil
+}
+
+func parsePromInstantValue(value []any) (float64, error) {
+	if len(value) < 2 {
+		return 0, fmt.Errorf("prometheus instant sample is missing a value")
+	}
+	switch v := value[1].(type) {
+	case string:
```
faiz9068, line 330: **A single `NaN` or `Inf` sample fails the whole call with an unactionable error.** `strconv.ParseFloat` accepts `"NaN"` and `"+Inf"` happily, and `encoding/json` cannot marshal either. Verified with a throwaway fixture carrying one `NaN` alongside one good row:

```
handler returned error: failed to marshal response: json: unsupported value: NaN
```

Every row is lost — including the healthy ones — and the message tells the caller nothing about which series was bad. It also quietly breaks the sort at `:152`: `vi != vj` is true for NaN-vs-NaN, so the `identityLess` tiebreak is skipped and a NaN row compares "greater" against everything, making the order non-deterministic in exactly the tool that promises determinism.

Fix: a `math.IsNaN` / `math.IsInf` check here — either skip the sample (a metric with no finite value isn't a rankable row) or return a named error identifying the class and series.

faiz9068 (commented):
Fresh full-range pass over `c3f7734..6baeee9` rather than another delta review — worth doing after five commits. Everything previously cleared still holds (aggregation, matcher anchoring, injection, the join, sort determinism, time arithmetic, no old-shape stragglers, no customer data). CI green; `go build` / `go vet` / `go test ./...` pass at `6baeee9`.

Three things earlier rounds missed, all three verified by execution. Two are inline; the third is below.

**`sum by (service_name, env)` double-counts dual-instrumented services.** `trace_endpoint_count` carries an `l9_source` label that the grouping drops, so a service observed by two instrumentation sources has both series summed into one `request_count`. Verified on a live deployment: 8 `(service, env)` pairs currently carry two `l9_source` groups, and for one of them the 10-minute split is `l9_source="ebpf"` → 1362 and the unlabelled SDK series → 2782 — the same inbound requests counted twice, ~1.5x inflation.

The inflation matters less than what it does to the *ordering*: dual-instrumented services outrank single-instrumented ones for free, on the tool being positioned as the deterministic fleet ranking. This is pre-existing — nothing in the repo touches `l9_source` — so I'd treat it as a caveat in `serviceSummaryCoverage` (`service_summary.go:28`) plus a follow-up ticket rather than something to solve here.

**Smaller items, no inline needed:**

- `ServiceSummaryArgs:66-73` still carries `jsonschema:"..."` tags that the explicit `InputSchema` now overrides — two copies of the same prose, free to drift. The `limit` contradiction inline below is exactly that drift having already happened.
- `dump_test.go:85` asserts only `len(enum) != 5`. Since `sortKeys` is derived from `serviceSummarySortSpecs`, the count cannot be wrong independently of the handler, so the assertion buys almost nothing — pin the actual key list if it's meant to be a contract test.
- `mergeServiceSummarySeries` overwrites on a duplicate `(service, env)` instead of accumulating. Unreachable today because `sum by` guarantees uniqueness, but it turns into a silent wrong answer the moment the grouping changes. `+=` or a one-line comment hardens it.
- An invalid `env` regex (e.g. `[`) surfaces as `400 Bad Request` with no mention of `env`. `escapePromQLLabel` closes the injection hole, but a `regexp.Compile` up front would give the model something actionable.
- `apm.go:523`, `apm.go:1065` and `apm.go:1512` still pass their PromQL-regex `env` into `BuildAPMServiceLink` as an exact catalog literal — the bug this PR fixed for `get_service_summary`. Now that `APMCatalogEnvFromRegex` exists they're one-line fixes, but they're out of scope here.

Nothing blocking beyond the two inline.

internal/apm/service_summary.go
prathamesh-sonpatki: Fixed in 13107e2: `mergeServiceSummarySeries` skips NaN/Inf samples instead of failing the whole call / poisoning sort. Added a unit test with mixed NaN/+Inf/finite rows.

prathamesh-sonpatki:
Addressed the full-range review in 13107e2:

**Inlines**
- Schema/limit: removed `minimum`/`maximum` so clamp stays reachable; description unchanged.
- NaN/Inf: non-finite samples are skipped (finite rows still returned).

**Also from the review body**
- `coverage` now caveats dual `l9_source` inflation.
- Invalid `env` regex rejected with an actionable error.
- Merge setters accumulate (`+=`) for duplicate `(service, env)`.
- `dump_test` pins the sort_by enum keys and asserts no schema limit bounds.
- Sibling APM deeplinks use `APMCatalogEnvFromRegex`; deviations uses `APMCatalogEnvExact`.

**Live verification** (built `bin/last9-mcp-pr210`, real org credentials):
- Handler: 10m window → 5 rows / matched=55 / truncated; `limit=150` clamps to 100; invalid `[` env errors on `env`.
- Dual `l9_source` evidence query: 6 `(service, env)` pairs with >1 source in this org.
- Local HTTP MCP (`LAST9_HTTP=true`) `tools/call` sort_by=`http_5xx_count` limit=3 returned ranked rows with `l9_source` coverage text.

internal/apm/service_summary_input_schema.go
prathamesh-sonpatki: Fixed in 13107e2: dropped schema `minimum`/`maximum` on `limit` so the handler clamp remains reachable for MCP clients (description stays accurate). `dump_test` now asserts bounds are absent and pins the sort_by key list.

internal/apm/apm.go
```diff
@@ -694,7 +520,7 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 
 		// Build deep link URL
 		dlBuilder := deeplink.NewBuilder(cfg.OrgSlug, cfg.ClusterID)
-		dashboardURL := dlBuilder.BuildAPMServiceLink(startTimeParam*1000, endTimeParam*1000, serviceName, env, "")
+		dashboardURL := dlBuilder.BuildAPMServiceLink(startTimeParam*1000, endTimeParam*1000, serviceName, deeplink.APMCatalogEnvFromRegex(env), "")
```
faiz9068: **This over-corrects and drops the deeplink env filter for the documented common input.** Same at `apm.go:1065` and `apm.go:1512`.

These three tools document `env` as an exact-ish name, not an anchored regex — `apm.go:53,61,68` all read *"Environment to filter by (default: .*, e.g. prod)"*. Nothing tells a model to send `^prod$`, so in practice every real call sends `prod`. Verified against `13107e2`:

```
env="prod"     FromRegex(apm.go)=false   Exact(deviations.go)=true
env="staging"  FromRegex(apm.go)=false   Exact(deviations.go)=true
env="^prod$"   FromRegex(apm.go)=true    Exact(deviations.go)=true
env=".*"       FromRegex(apm.go)=false   Exact(deviations.go)=false
```

At `6baeee9`, `BuildAPMServiceLink` applied `APMCatalogEnvExact` internally, so `prod` did produce a filter. As of this commit `get_service_performance_details`, `get_service_operations_summary` and `get_service_dependency_graph` silently lose it on essentially every call.

`deviations.go:199` got `APMCatalogEnvExact` in this same commit, which is the right choice for a caller whose env is a plain name — these three should match it. That also keeps `.*` screened out, which was the original concern.

My fault for how I framed the earlier finding: I flagged that these sites passed a regex through as a literal, and the correction landed one notch too strict.

faiz9068 (commented):
Re-reviewed `6baeee9..13107e2`. All five items are fixed and I verified each: `limit` bounds dropped from the schema so the clamp is reachable and the prose is true again (with `dump_test.go` inverted to assert their absence), non-finite samples skipped with both reasons in the comment, `+=` accumulation with a direct unit test, `env` regex validated up front, and the `l9_source` caveat in both `coverage` and the description. `go test ./...` passes locally at `13107e2`.

One regression introduced by the deeplink rewiring — inline.

**CI note:** `Analyze (javascript-typescript)` is red and `CodeQL` neutral on this head, but it's infrastructure, not code — `An action could not be found at the URI 'https://codeload.github.com/actions/setup-go/tar.gz/924ae3a…'`. The same checks passed on all four earlier commits; a re-run should clear it.

internal/apm/apm.go
prathamesh-sonpatki: Fixed in the follow-up commit: the three sibling tools now use `APMCatalogEnvExact` (same as deviations), so plain `prod`/`staging` keep their catalog filter while `.*` stays screened out. `get_service_summary` still uses `APMCatalogEnvFromRegex` because its `env` is documented as a PromQL regex.

faiz9068 (approved):

PR #212, opened 2026-08-14
fix(logs): align get_logs whale + fail-closed pipeline validation

prathamesh-sonpatki:
## Summary
- Align `get_logs` whale `window_aggregate` with last9/api (`function`/`as`/`window`) and add NOT-SQL / bare-field / order rules (ENG-1730, surgical ENG-1452/1410).
- Fail-closed pipeline validation before upstream or attrs fan-out; typed `LogPipelineValidationError`; `$and` wrap; thin `get_logs` InputSchema (ENG-1729 MVP + thin ENG-953).
- Fix traces lookback docs 5→60 (and matching `tracejson.md`); delete unused description stubs; point empty-query errors at `last9://reference/*`.

## Test plan
- [x] `go test ./internal/prompts/ ./internal/telemetry/logs/ ./ -count=1`
- [x] `dump-tools` vs main: candidate has no `window_minutes`, has NOT SQL + InputSchema oneOf
- [x] Live MCP: canonical WA succeeds; `window_minutes` / SpanKind / bad parse reject locally
- [x] Companion evals: https://github.com/last9/last9-mcp-evals/pull/49 — 50/50 on candidate; WA cases fail on main

Linear: ENG-1730, ENG-1729 (MVP), ENG-1452 (surgical), ENG-953 (thin get_logs), ENG-1410

**Deferred:** stage ordering (filter→parse→aggregate) is documented in the whale/`logjson` manual but not enforced by `validateLogJSONQuery` (MVP).

**Review follow-up (`8d53762`):** tip-reachable schema (`anyOf`, no stage `additionalProperties:false`, loose parse/WA `required`), allow TraceId/SpanId/ParentSpanId, default parse `field`→Body, dotted labels via `safeBodyKeyPattern`, typed sanitize errors, CHANGELOG + whale budget tighten.

### DIFF
--- a/dump_test.go
+++ b/dump_test.go
@@ -61,8 +61,14 @@ func TestDumpTools(t *testing.T) {
 		t.Fatal("get_logs description still contains unsubstituted {{labels}} placeholder")
 	}
 	for _, whale := range []string{"get_logs", "get_traces", "get_service_logs"} {
-		if len(out.Tools[byName[whale]].Description) > 2000 {
-			t.Fatalf("%s served description length %d exceeds 2000-char tripwire", whale, len(out.Tools[byName[whale]].Description))
+		whaleBudgets := map[string]int{
+			"get_logs":         4500,
+			"get_traces":       2200,
+			"get_service_logs": 2000,
+		}
+		budget := whaleBudgets[whale]
+		if len(out.Tools[byName[whale]].Description) > budget {
+			t.Fatalf("%s served description length %d exceeds %d-char budget", whale, len(out.Tools[byName[whale]].Description), budget)
 		}
 		if !strings.Contains(out.Tools[byName[whale]].Description, "last9://reference/") {
 			t.Fatalf("%s description missing resource URI pointer", whale)
--- a/dump_test.go
+++ b/dump_test.go
@@ -60,12 +60,12 @@ func TestDumpTools(t *testing.T) {
 	if strings.Contains(out.Tools[byName["get_logs"]].Description, "{{labels}}") {
 		t.Fatal("get_logs description still contains unsubstituted {{labels}} placeholder")
 	}
+	whaleBudgets := map[string]int{
+		"get_logs":         2600,
+		"get_traces":       2200,
+		"get_service_logs": 2000,
+	}
 	for _, whale := range []string{"get_logs", "get_traces", "get_service_logs"} {
-		whaleBudgets := map[string]int{
-			"get_logs":         4500,
-			"get_traces":       2200,
-			"get_service_logs": 2000,
-		}
 		budget := whaleBudgets[whale]
 		if len(out.Tools[byName[whale]].Description) > budget {
 			t.Fatalf("%s served description length %d exceeds %d-char budget", whale, len(out.Tools[byName[whale]].Description), budget)
--- a/dump_test.go
+++ b/dump_test.go
@@ -105,12 +105,29 @@ func TestDumpTools(t *testing.T) {
 	if strings.Contains(out.Tools[byName["get_logs"]].Description, "{{labels}}") {
 		t.Fatal("get_logs description still contains unsubstituted {{labels}} placeholder")
 	}
+	// get_logs: budget covers combined (description + inputSchema JSON) served size
+	// because get_logs ships a hand-crafted schema that contributes meaningfully to
+	// the context window. Budget is 7000 with headroom over the baseline ~6283.
+	logsIdx := byName["get_logs"]
+	logsSchemaBytes, err := json.Marshal(out.Tools[logsIdx].InputSchema)
+	if err != nil {
+		t.Fatalf("marshal get_logs inputSchema: %v", err)
+	}
+	logsServedSize := len(out.Tools[logsIdx].Description) + len(logsSchemaBytes)
+	const logsServedBudget = 7000
+	if logsServedSize > logsServedBudget {
+		t.Fatalf("get_logs combined (description + inputSchema) served size %d exceeds %d-char budget", logsServedSize, logsServedBudget)
+	}
+	if !strings.Contains(out.Tools[logsIdx].Description, "last9://reference/") {
+		t.Fatal("get_logs description missing resource URI pointer")
+	}
+
+	// Description-only budgets for other whales (main tripwire values restored to 2000).
 	whaleBudgets := map[string]int{
-		"get_logs":         2600,
-		"get_traces":       2200,
+		"get_traces":       2000,
 		"get_service_logs": 2000,
 	}
-	for _, whale := range []string{"get_logs", "get_traces", "get_service_logs"} {
+	for _, whale := range []string{"get_traces", "get_service_logs"} {
 		budget := whaleBudgets[whale]
 		if len(out.Tools[byName[whale]].Description) > budget {
 			t.Fatalf("%s served description length %d exceeds %d-char budget", whale, len(out.Tools[byName[whale]].Description), budget)
--- a/internal/prompts/descriptions/get_logs_base.md
+++ b/internal/prompts/descriptions/get_logs_base.md
@@ -1,21 +1,28 @@
-Query logs with `logjson_query` — JSON **array of stages**. Each stage `"type"`: `filter`|`parse`|`aggregate`|`window_aggregate`. No `"stage"`/`"conditions"`.
+`logjson_query` is a JSON **array of stages** — NOT SQL, NOT a query string. Each stage sets `"type"`: `filter`|`parse`|`aggregate`|`window_aggregate`. No `"stage"`/`"conditions"`.
 
-**Filter:** `{"type":"filter","query":{"$and":[{"$eq":["SeverityText","ERROR"]}]}}` — `$eq|$neq|$contains|$gt|$gte|$lt|$lte|$regex` on `[field, value]` strings. Always `$and`-wrap.
+**Order rule:** filter first → parse (before using extracted attrs) → aggregate or window_aggregate only for counts/trends. Never use SpanKind/StatusCode/Duration/SpanName as log filters.
 
-**Parse / aggregate:** parse Body (`json`/`logfmt`/`regexp`), then filter `attributes['…']`. **Aggregate:** `{"type":"aggregate","aggregates":[{"function":{"$count":[]},"as":"_count"}],"groupby":{"ServiceName":"service"}}` — `function` uses `{"$count":[]}`, `{"$max":["field"]}`, or `{"$avg":["field"]}`. **window_aggregate** for trends/per-minute counts — `aggregates`+`window_minutes`, not `TimeBucket`.
+**Filter:** `{"type":"filter","query":{"$and":[{"$eq":["SeverityText","ERROR"]}]}}` — operators: `$eq|$neq|$ieq|$contains|$containsWords|$gt|$gte|$lt|$lte|$regex`. Always `$and`-wrap. Values are strings.
 
-**Severity-less logs:** empty `SeverityText` → parse Body `level`, gate `$eq`/`$ieq` on `ERROR` before counting.
+**Parse:** `{"type":"parse","parser":"json","field":"Body","labels":{"key":"key"}}` — `parser` is `json`/`logfmt`/`regexp` (never `"format"`). Parse Body first; then filter on `attributes['key']`.
 
-**Existence / attrs:** exists → `{"$neq":["field",""]}` (never `$exists`). Structured fields → `attributes['key']` — not Body `$contains`.
+**Aggregate:** `{"type":"aggregate","aggregates":[{"function":{"$count":[]},"as":"_count"}],"groupby":{"ServiceName":"service"}}` — `aggregates` (plural), `function` as object.
+
+**window_aggregate** for per-minute/trend counts — use `function`+`as`+`window` (NOT the aggregate stage, NOT `TimeBucket`):
+`[{"type":"filter","query":{"$and":[{"$eq":["SeverityText","ERROR"]}]}},{"type":"window_aggregate","function":{"$count":[]},"as":"errors","window":["1","minutes"]}]`
+
+**Severity-less logs:** empty `SeverityText` → parse Body `level`, gate `$ieq` on `ERROR` before counting.
+
+**Existence / attrs:** exists → `{"$neq":["field",""]}` (never `$exists`). Never bare field names — dotted (`http.status_code`) OR single-token (`community_member_id`) — wrap as `attributes['key']`/`resources['key']`. Only top-level `ServiceName`/`Body`/`SeverityText`/`Timestamp` may be bare.
 
 **Scope:** tenant → `resources['last9.tenant']`; env → `resources['deployment.environment']`. User `service.name` → `ServiceName`; `k8s.*` → `resources['k8s.…']`.
 
 **Free-text IDs** (EPL_…) → `{"$contains":["Body","…"]}` — not `ServiceName`.
 
-**HTTP 5xx:** filter status field with literal code (e.g. `$eq` on `attributes['status_code']`,`"500"`) — never `SeverityText`/`ERROR`; avoid regex-only when user names a code.
+**HTTP 5xx:** filter status field with literal code (e.g. `$eq` on `attributes['status_code']`, `"500"`) — never `SeverityText`/`ERROR`; avoid regex-only when user names a code.
 
-**Time:** `lookback_minutes` (default **5**). **Absolute ISO bounds** → `start_time_iso`+`end_time_iso` on the tool call — never `Timestamp`/`$gte`/`$lte` in the pipeline (pipeline filters log data only).
+**Time:** `lookback_minutes` (default **5**). **Absolute ISO bounds** → `start_time_iso`+`end_time_iso` on the tool call — never `Timestamp`/`$gte`/`$lte` in the pipeline.
 
-**l9_sanity:** high `ratio` or "filter likely too broad" → next call `get_logs` with `aggregate`+`$count` and an ERROR/`SeverityText` gate — not discovery-only.
+**l9_sanity:** high `ratio` or "filter likely too broad" → next call `get_logs` with `aggregate`+`$count` and an ERROR/`SeverityText` gate.
 
 Full manual: `last9://reference/logjson`
--- a/internal/prompts/descriptions/get_traces_base.md
+++ b/internal/prompts/descriptions/get_traces_base.md
@@ -12,7 +12,7 @@ Query traces with `tracejson_query` — a JSON **array of stages**. Each stage M
 
 **Scope:** tenant name → `resources['last9.tenant']`; deployment env → `resources['deployment.environment']`.
 
-**Time (tool args):** Prefer `lookback_minutes` (default **5**). Absolute → `start_time_iso`+`end_time_iso` (RFC3339). Never put the window as a `Timestamp` filter in the pipeline.
+**Time (tool args):** Prefer `lookback_minutes` (default **60**). Absolute → `start_time_iso`+`end_time_iso` (RFC3339). Never put the window as a `Timestamp` filter in the pipeline.
 
 **Fields:** Top-level: `TraceId`, `SpanId`, `ServiceName`, `SpanName`, `SpanKind`, `StatusCode`, `Duration`, `Timestamp`, `ParentSpanId`. SpanKind/StatusCode need full OTel prefixes (`SPAN_KIND_SERVER`, `STATUS_CODE_ERROR`—not `SERVER`/`ERROR`). **`Duration` is nanoseconds** (1000ms = `1000000000`). Span/resource attrs → `get_trace_attributes*`; use `attributes['key']` / `resources['key']` (never `SpanAttributes.foo`).
 
--- a/internal/prompts/prompts_test.go
+++ b/internal/prompts/prompts_test.go
@@ -53,7 +53,7 @@ func TestGetTracesDescriptionCriticalRules(t *testing.T) {
 		{"groupby", "must document groupby key name"},
 		{"resources['last9.tenant']", "must document tenant scoping"},
 		{"last9://reference/tracejson", "must point at the tracejson resource"},
-		{"60", "must document default lookback of 60 minutes"},
+		{"default **60**", "must document default lookback of 60 minutes"},
 	}
 	for _, c := range checks {
 		if !strings.Contains(desc, c.phrase) {
@@ -85,7 +85,7 @@ func TestGetServiceLogsDescriptionCriticalRules(t *testing.T) {
 
 func TestWhaleDescriptionsBounded(t *testing.T) {
 	budgets := map[string]int{
-		"get_logs_base":         4500,
+		"get_logs_base":         2600,
 		"get_traces_base":       2200,
 		"get_service_logs_base": 2000,
 	}
--- a/internal/prompts/prompts_test.go
+++ b/internal/prompts/prompts_test.go
@@ -17,16 +17,24 @@ func TestGetLogsDescriptionCriticalRules(t *testing.T) {
 		reason string
 	}{
 		{"window_aggregate", "must document time-bucketed counts"},
+		{"window", "must document window key for window_aggregate"},
+		{"function", "must document function key for window_aggregate"},
+		{"NOT SQL", "must clarify logjson_query is not SQL"},
 		{"$neq", "must document existence idiom"},
 		{"start_time_iso", "must document absolute time on tool args"},
 		{"resources['last9.tenant']", "must document tenant scoping"},
 		{"resources['deployment.environment']", "must document env scoping"},
 		{"last9://reference/logjson", "must point at the logjson resource"},
+		{"community_member_id", "must document bare single-token field mistake"},
+		{"attributes['key']", "must require attributes/resources wrapping"},
 	}
 	for _, c := range checks {
 		if !strings.Contains(desc, c.phrase) {
 			t.Errorf("GetLogsDescription missing %q: %s", c.phrase, c.reason)
 		}
+	}
+	if strings.Contains(desc, "window_minutes") {
+		t.Error("GetLogsDescription must NOT contain deprecated 'window_minutes' key")
 	}
 }
 
@@ -45,6 +53,7 @@ func TestGetTracesDescriptionCriticalRules(t *testing.T) {
 		{"groupby", "must document groupby key name"},
 		{"resources['last9.tenant']", "must document tenant scoping"},
 		{"last9://reference/tracejson", "must point at the tracejson resource"},
+		{"60", "must document default lookback of 60 minutes"},
 	}
 	for _, c := range checks {
 		if !strings.Contains(desc, c.phrase) {
@@ -75,16 +84,23 @@ func TestGetServiceLogsDescriptionCriticalRules(t *testing.T) {
 }
 
 func TestWhaleDescriptionsBounded(t *testing.T) {
-	for name, body := range map[string]string{
+	budgets := map[string]int{
+		"get_logs_base":         4500,
+		"get_traces_base":       2200,
+		"get_service_logs_base": 2000,
+	}
+	bodies := map[string]string{
 		"get_logs_base":         prompts.GetLogsDescription,
 		"get_traces_base":       prompts.GetTracesDescription,
 		"get_service_logs_base": prompts.GetServiceLogsDescription,
-	} {
+	}
+	for name, body := range bodies {
 		if len(body) == 0 {
 			t.Errorf("%s empty", name)
 		}
-		if len(body) > 2000 {
-			t.Errorf("%s length %d exceeds 2000-char tripwire", name, len(body))
+		budget := budgets[name]
+		if len(body) > budget {
+			t.Errorf("%s length %d exceeds %d-char budget", name, len(body), budget)
 		}
 	}
 }
--- a/internal/prompts/prompts_test.go
+++ b/internal/prompts/prompts_test.go
@@ -86,7 +86,7 @@ func TestGetServiceLogsDescriptionCriticalRules(t *testing.T) {
 func TestWhaleDescriptionsBounded(t *testing.T) {
 	budgets := map[string]int{
 		"get_logs_base":         2600,
-		"get_traces_base":       2200,
+		"get_traces_base":       2000,
 		"get_service_logs_base": 2000,
 	}
 	bodies := map[string]string{
--- a/internal/telemetry/logs/input_schema.go
+++ b/internal/telemetry/logs/input_schema.go
@@ -60,7 +60,9 @@ func logFilterStageSchema() map[string]interface{} {
 	return map[string]interface{}{
 		"type":        "object",
 		"description": "Filter stage: narrows the dataset by condition. Always wrap in $and.",
-		"required":    []string{"type", "query"},
+		// Only "type" is schema-required; validateLogJSONQuery owns "query" tips so
+		// {"type":"filter","conditions":{...}} mistakes reach the conditions tip.
+		"required": []string{"type"},
 		"properties": map[string]interface{}{
 			"type": logStringEnum("filter"),
 			"query": map[string]interface{}{
@@ -94,10 +96,11 @@ func logParseStageSchema() map[string]interface{} {
 }
 
 func logAggregateStageSchema() map[string]interface{} {
+	// No required on aggregateFuncSchema; validateLogJSONQuery owns function/as/alias
+	// tips so {"function":{...},"alias":"count"} mistakes reach the alias tip.
 	aggregateFuncSchema := map[string]interface{}{
 		"type":        "object",
 		"description": "Aggregate entry. Use 'aggregates' (plural) and 'as'. Function MUST be an object like {\"$count\": []} — never a string.",
-		"required":    []string{"function", "as"},
 		"properties": map[string]interface{}{
 			"function": map[string]interface{}{
 				"type":        "object",
@@ -113,7 +116,9 @@ func logAggregateStageSchema() map[string]interface{} {
 	return map[string]interface{}{
 		"type":        "object",
 		"description": "Aggregate stage. Use 'aggregates' (NOT 'aggregations') and 'groupby' (NOT 'group_by').",
-		"required":    []string{"type", "aggregates"},
+		// Only "type" is schema-required; validateLogJSONQuery owns "aggregates" tips
+		// so {"type":"aggregate","aggs":[...]} mistakes reach the aggs tip.
+		"required": []string{"type"},
 		"properties": map[string]interface{}{
 			"type": logStringEnum("aggregate"),
 			"aggregates": map[string]interface{}{
--- a/internal/telemetry/logs/input_schema.go
+++ b/internal/telemetry/logs/input_schema.go
@@ -41,6 +41,12 @@ func GetLogsInputSchema() map[string]interface{} {
 	}
 }
 
+// logjsonQuerySchema returns the JSON Schema for the logjson_query array.
+// The anyOf branches are descriptive guidance — they do NOT own rejection.
+// The permissive catch-all at the end ensures that wrong types, unknown enums,
+// and partial stages (e.g. SQL strings as query, type "sort", parser "grok",
+// function as string) pass schema validation and reach validateLogJSONQuery,
+// which returns actionable tip errors instead of opaque anyOf failures.
 func logjsonQuerySchema() map[string]interface{} {
 	return map[string]interface{}{
 		"type":        "array",
@@ -51,6 +57,7 @@ func logjsonQuerySchema() map[string]interface{} {
 				logParseStageSchema(),
 				logAggregateStageSchema(),
 				logWindowAggregateStageSchema(),
+				map[string]interface{}{"type": "object"}, // catch-all: wrong types/enums reach validate tips
 			},
 		},
 	}
--- a/internal/telemetry/logs/input_schema.go
+++ b/internal/telemetry/logs/input_schema.go
@@ -6,8 +6,10 @@ func logStringEnum(values ...string) map[string]interface{} {
 
 // GetLogsInputSchema returns a hand-crafted JSON Schema for the get_logs tool.
 // This replaces the auto-generated schema from GetLogsArgs so that logjson_query
-// items have a detailed oneOf definition — constraining the LLM to use correct
-// stage shapes and preventing wrong-key mistakes (window_minutes, format, etc.).
+// items have a detailed anyOf definition guiding stage shapes. Schema stays
+// deliberately loose on stage required/additionalProperties so wrong-key
+// mistakes (window_minutes, format, …) reach validateLogJSONQuery tip errors
+// instead of an opaque anyOf failure from the SDK.
 func GetLogsInputSchema() map[string]interface{} {
 	return map[string]interface{}{
 		"type": "object",
@@ -34,7 +36,7 @@ func GetLogsInputSchema() map[string]interface{} {
 				"description": "Optional log index in the form physical_index:<name> or rehydration_index:<block_name>. Omit when the user did not specify an index.",
 			},
 		},
-		"required":              []string{"logjson_query"},
+		"required":             []string{"logjson_query"},
 		"additionalProperties": false,
 	}
 }
@@ -44,7 +46,7 @@ func logjsonQuerySchema() map[string]interface{} {
 		"type":        "array",
 		"description": "JSON pipeline query for logs. An ordered list of stages: filter → parse → aggregate/window_aggregate. NOT SQL, NOT a query string. Each stage is an object with a 'type' field.",
 		"items": map[string]interface{}{
-			"oneOf": []interface{}{
+			"anyOf": []interface{}{
 				logFilterStageSchema(),
 				logParseStageSchema(),
 				logAggregateStageSchema(),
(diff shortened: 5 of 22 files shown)


prathamesh-sonpatki:
Follow-up commit on this PR from latest hosted `get_logs` ERROR review (24h, last9-mcp):

- require parse `field` (usually Body)
- coerce numeric/bool filter values → strings (`500` → `"500"`)
- reject `SpanName`/other trace-only fields in groupby (already rejected in filters)
- reject bad regexp named captures (`(?P<$merchant>`) and `$…` label keys

These cover the residual `invalid JSON pipeline` / bad-regexp classes that whale+WA/`format`/SpanKind alone would not fully stop.

internal/telemetry/logs/input_schema.go
```diff
@@ -0,0 +1,156 @@
...
+		"description": "Parse stage: extracts fields from Body using json, logfmt, or regexp parsers. Never use 'format' — key is 'parser'.",
+		"required":    []string{"type", "parser"},
+		"properties": map[string]interface{}{
+			"type":    logStringEnum("parse"),
+			"parser":  logStringEnum("json", "logfmt", "regexp"),
+			"field":   map[string]interface{}{"type": "string", "description": "Field to parse (typically Body)."},
+			"pattern": map[string]interface{}{"type": "string", "description": "Regexp pattern with named capture groups (regexp parser only)."},
+			"labels":  map[string]interface{}{"type": "object", "description": "Field mappings for json/logfmt parsing."},
+		},
+		"additionalProperties": false,
```
faiz9068: **Critical — this `additionalProperties: false` makes the PR's headline error messages unreachable.**

`registerIfAllowed` → `RegisterInstrumentedTool` → `mcp.AddTool`. In go-sdk v1.5.0 `setSchema` remarshals a provided `map[string]any` into `*jsonschema.Schema` and resolves it, and the generated handler runs `applySchema(input, inputResolved)` **before** unmarshalling into `GetLogsArgs` and before your handler is entered. So this schema is enforced at the protocol layer, ahead of `prepareLogJSONQuery`.

Because parse/aggregate/window_aggregate all combine `additionalProperties: false` with a pinned `type` enum, the three mistakes this PR was written to fix fail *all four* `oneOf` branches. I resolved the served schema and ran them through it — for `format` on parse, and for `window_minutes` / `aggregates` on window_aggregate, the model receives exactly:

```
validating "arguments": validating root: validating /properties/logjson_query:
validating /properties/logjson_query/items: oneOf: did not validate against any of
[<anonymous schema> <anonymous schema> <anonymous schema> <anonymous schema>]
```

No stage index, no key name, no hint. Meanwhile the carefully written tip in `query_validate.go` (`common mistake: window_aggregate uses "function"+"as"+"window" not "aggregates" or "window_minutes"`) is unreachable for every stage type except `filter` — the only branch here without `additionalProperties: false`. `TestUnknownKeyTipsAreStageSpecific` passes because it calls `prepareLogJSONQuery` directly, which is a path the served tool never reaches for these inputs.

This inverts the goal: pre-PR the model got an upstream 400 with a decent server-side hint appended; post-PR, for the flagship mistakes, it gets a generic `oneOf` failure it cannot act on.

Suggest dropping `additionalProperties: false` from all three stage branches (keep it at the root on line 38, where it does produce a readable unknown-key error) and letting `validateLogJSONQuery` own key rejection, since that's where the good messages live. Keep the `type` enums — those are what make `oneOf` discriminate, and they don't fire on unknown keys. Same applies to lines 108, 126 and 154.

internal/telemetry/logs/input_schema.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+
+func logjsonQuerySchema() map[string]interface{} {
+	return map[string]interface{}{
+		"type":        "array",
+		"description": "JSON pipeline query for logs. An ordered list of stages: filter → parse → aggregate/window_aggregate. NOT SQL, NOT a query string. Each stage is an object with a 'type' field.",
+		"items": map[string]interface{}{
+			"oneOf": []interface{}{
```
faiz9068: Prefer `anyOf` over `oneOf` here. The four branches are mutually exclusive by construction (the `type` enum discriminates), so `anyOf` is functionally identical — but OpenAI's structured-output subset supports `anyOf` and not `oneOf`. `schema_test.go:54-59` shows this repo already cares about strict OpenAI-side validators (that's why `items` must not be boolean), and this is a served tool schema on a public MCP server, so it's an API-compat question rather than a style one.

internal/telemetry/logs/input_schema.go
```diff
@@ -0,0 +1,156 @@
...
+			},
+		},
+	}
+}
+
+func logParseStageSchema() map[string]interface{} {
+	return map[string]interface{}{
+		"type":        "object",
+		"description": "Parse stage: extracts fields from Body using json, logfmt, or regexp parsers. Never use 'format' — key is 'parser'.",
+		"required":    []string{"type", "parser"},
```
faiz9068: This says `field` is optional on `parse`, while `query_validate.go:247` hard-rejects a parse stage without it. Both can't be right — see my comment on that line; the API and `references/logjson.md` side with this one.

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,445 @@
...
+		Category: category,
+		Path:     path,
+		Message:  message,
+	}
+}
+
+// traceOnlyLogFields are span/trace fields that have no meaning as top-level
+// log filters. Accepting them silently produces zero results; rejecting them
+// surfaces the model error before the API call.
+var traceOnlyLogFields = map[string]struct{}{
```
faiz9068, line 54: **Critical — `TraceId` / `SpanId` / `ParentSpanId` are populated log fields, and rejecting them breaks log↔trace correlation.**

I verified against the live logjson API. A pipeline of `{"type":"filter","query":{"$and":[{"$neq":["TraceId",""]}]}}` followed by a `$count` aggregate returns thousands of matching log rows over a 15-minute window — no 400, real data. So the rejection message ("this field does not exist in logs and will match nothing") states something untrue, and a model that reads it will not retry: it will conclude the correlation is impossible. "Show me the logs for this trace" is a core workflow and it's now unreachable through `get_logs`, both as a filter and as a `groupby`.

The strictness itself is well-founded — I confirmed `SeverityNumber` genuinely 400s upstream — the set is just too broad. Suggest dropping `TraceId`, `SpanId`, `ParentSpanId` here and adding them to `isCanonicalLogFieldRef`'s bare-allowed set, keeping `SpanKind`, `StatusCode`, `Duration`, `SpanName`. Note the two lists have to move together: this map currently does double duty as both "reject in filter/groupby" (lines 317, 408) and "reject as a field ref" (via `isTraceOnlyLogField` in `query_sanitize.go:379`).

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,445 @@
...
+			fmt.Sprintf("invalid parser %q at %s — must be one of: json, logfmt, regexp (never \"format\")", parser, stagePath),
+		)
+	}
+
+	field, _ := stage["field"].(string)
+	if strings.TrimSpace(field) == "" {
+		return newLogValidationError(
+			LogValidationMissingRequired,
+			stagePath+".field",
+			fmt.Sprintf("parse stage at %s missing required \"field\" key — usually \"field\":\"Body\"", stagePath),
```
faiz9068: **Critical — `field` is not actually required, and this rejects the repo's own documented examples.**

Three-way disagreement:

1. **Upstream accepts it.** A `{"type":"parse","parser":"json"}` stage with no `field`, between a filter and a `$count` aggregate, returns a result — no 400.
2. **`references/logjson.md` documents parse without `field`** in three places: the stage spec (lines 197-203: `type`, `parser`, `pattern`, `labels`), Example 5 (line 411) and Example 7 (line 531). That's the resource the tool description points the model at. Running those two documented examples through `prepareLogJSONQuery` gives `REJECTED: parse stage at logjson_query[1] missing required "field" key`.
3. **`input_schema.go:81`** lists only `["type","parser"]` as required.

Either default it (`if field == "" { stage["field"] = "Body" }` in the sanitize step — normalizing beats rejecting when the correct value is knowable) or keep it required and update `logjson.md` + `input_schema.go` in the same commit. The current state can't ship as-is.

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,445 @@
...
+		)
+	}
+
+	if labels, ok := stage["labels"].(map[string]interface{}); ok {
+		for labelKey := range labels {
+			if !validNamedCaptureName.MatchString(labelKey) {
```
faiz9068: **Critical — this rejects the exact parse stage the server's own discovery tool emits.**

`bodyDerivedHint` (`attributes_for_pipeline.go:68-78`) emits `"labels": {key: key}` where `key` passed `safeBodyKeyPattern` (line 64) — which explicitly allows `.`, `@`, `/`, `-`. So `get_log_attributes_for_pipeline` hands the model:

```json
[{"type":"parse","parser":"json","field":"Body","labels":{"http.status_code":"http.status_code"}}, ...]
```

and `get_logs` then rejects it with `parse labels key "http.status_code" ... never "$merchant" or dotted names`. I confirmed upstream accepts dotted label keys. Dotted JSON body keys are the common case (`http.status_code`, `user.id`), and `logjson.md:45` names `http.status_code` as *the* motivating example for using discovery — so the primary body-parse workflow is broken for precisely the keys discovery exists to find.

Root cause is scope: `validNamedCaptureName` is correct for **Go regexp named captures** (`(?P<name>)` genuinely forbids dots), but it's been generalized to json/logfmt `labels`, where the alphabet constraint doesn't apply. Suggest keeping `validNamedCaptureName` to `validateRegexpNamedCaptures` (line 293) only, and validating `labels` keys against something like `safeBodyKeyPattern` — rejecting quotes/spaces/backslashes/`$` so the `$merchant` case this was written for still gets caught, while allowing dots.

A round-trip test would pin this permanently: feed `bodyDerivedHint` / `logfmtBodyDerivedHint` / `regexpBodyDerivedHint` output (including a dotted key) straight into `prepareLogJSONQuery` and assert acceptance. A server rejecting its own hint is about the worst failure mode an MCP tool has, and it's cheap to prevent.

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,445 @@
...
+
+// validNamedCaptureName is the Go regexp named-group alphabet (no `$`, hyphens, etc.).
+var validNamedCaptureName = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
+
+// Validation error categories for fail-closed log pipeline checks.
+// Stable strings — safe for metrics labels; never include customer payloads.
+const (
+	LogValidationUnknownStageType = "unknown_stage_type"
+	LogValidationUnknownStageKey  = "unknown_stage_key"
+	LogValidationMissingRequired     = "missing_required"
```
faiz9068: The typed error only covers half the fail-closed surface. `LogPipelineValidationError` is returned only from `validateLogJSONQuery`; every rejection from the sanitize layer returns a bare `fmt.Errorf` — bare field names (`query_sanitize.go:352`), trace-only field refs (`:341`), double-quoted brackets (`:331`), `resource_` prefix (`:346`), groupby collisions (`:305`), non-string filter values (`:213`). In practice that's the majority of rejections.

And `Category` / `Path` have zero non-test consumers (I grepped) despite the doc comment promising "stable strings — safe for metrics labels". Either wire the taxonomy through the sanitize layer and actually emit the metric, or drop the struct and keep plain errors — right now it's an abstraction that exists to satisfy two tests.

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,445 @@
...
+func stageKeyList(allowed map[string]struct{}) string {
+	keys := make([]string, 0, len(allowed))
+	for k := range allowed {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return strings.Join(keys, ", ")
+}
+
+func windowLen(v interface{}) int {
```
faiz9068: Minor: `windowLen` is redundant. At its only call site (line 370) the `[]interface{}` assertion has already been made and `ok` already distinguishes wrong-type from wrong-length — so `len(window)` guarded on `ok` does the same job without the helper.

Also, both this file and `input_schema.go` are unformatted — `gofmt -l` flags them (const block alignment around lines 18-24 here, `"required"` alignment at `input_schema.go:37`). `query_sanitize.go` was already unformatted at base, so that one isn't yours.

internal/telemetry/logs/schema_test.go
```diff
@@ -146,3 +146,129 @@ func TestGetLogsArgs_LookbackDescriptionMatchesPromptDefault(t *testing.T) {
...
+	if !found {
+		t.Error("window_aggregate stage schema not found in oneOf")
+	}
+
+	if schema["additionalProperties"] != false {
+		t.Error("root GetLogsInputSchema must set additionalProperties: false")
+	}
+}
+
+func TestGetLogsInputSchema_RejectsUnknownTopLevelKeys(t *testing.T) {
```
faiz9068, line 253: This is the right shape of test — resolving through `jsonschema.Schema.Resolve` and validating a real document is what proves the hand-written schema is well-formed, rather than asserting on map keys. Two gaps though:

1. It asserts only `err != nil`, so it would still pass if the `oneOf` were unsatisfiable for the accompanying filter stage.
2. There's no positive test that a *valid* pipeline of each of the four stage types passes `resolved.Validate`. I checked by hand that all four stage types and all six top-level params (including explicit `null`s) do validate — but nothing here would catch a regression, and for an MCP tool an unsatisfiable schema means the model can't query logs at all.

A small table of one valid document per stage type would close both.

faiz9068 (commented):
Reviewed at `5555f5e`. Verification I ran locally in a throwaway worktree: `go build ./...` and `go vet ./...` clean, `go test ./... -count=1` passes all 17 packages, `go run . dump-tools` serves and round-trips the new hand-written `get_logs` schema. The test claim in the description holds — the findings below are things the suite doesn't cover, and several were settled by probing the live logjson API.

**The fail-closed claim is structurally true**, and I want to say that first because it's the load-bearing part. There are exactly two entry points that accept a caller-supplied pipeline (`GetLogsArgs.LogjsonQuery`, `GetLogAttributesForPipelineArgs.Pipeline`) and both call `prepareLogJSONQuery` before anything else. In `attributes_for_pipeline.go` validation lands before the `params` build, before `fetchLogSeriesFieldNames`, and before the `sampleBodyDerivedAttributes` goroutine is spawned — so the attrs fan-out genuinely cannot fire on an invalid pipeline. `get_service_logs` builds its pipeline from typed args and is correctly out of scope. I looked for a bypass and didn't find one. `TestGetLogAttributesForPipeline_InvalidPipelineRejectsBeforeHTTP` asserting `requestCount == 0` against a real `httptest` server is the right way to prove that. The error-message craft is also genuinely good — path-scoped, states the correct form, points at `get_log_attributes` for self-correction. That's the hard part of writing a validator whose consumer is a model, and it was done well.

What blocks merge is that the validator over-rejects in four verified ways, and for an MCP tool a false positive is indistinguishable from "broken" — the model reads the rejection, believes it, and stops. Details inline; the two structural ones:

- The hand-written `oneOf` schema is enforced by the SDK **ahead of** the handler, so the tailored `window_minutes` / `format` / `aggregates` self-correction messages this PR exists to deliver are unreachable. The model instead gets a bare `oneOf: did not validate against any of [...]`.
- Three places now encode the pipeline grammar — `input_schema.go`, `validateLogJSONQuery`, and `references/logjson.md` — with nothing keeping them in sync, and they already disagree on whether `parse.field` is required. Worth picking one as source of truth (I'd argue the Go validator, since it's the only one that can produce a self-correcting message), making the JSON Schema deliberately looser, and adding a test that asserts every parse/aggregate/window_aggregate example extracted from `logjson.md` survives `prepareLogJSONQuery`. That single test catches two of the four Criticals.

Two things not anchorable to a diff line:

- **`references/tracejson.md` now contradicts the description it backs.** Lines 38, 42 and 447 still say `lookback_minutes: 5`, and 447 is emphatic: "CRITICAL: When no time is specified, MUST use lookback_minutes: 5 (NOT 60!)". Both files are served to the model and the manual is the louder one, so fixing the short description alone makes the contradiction worse than leaving it. (Also the PR body says this fix was 24h→60; the diff is 5→60, and there's no "24h" anywhere — worth correcting so reviewers aren't hunting for a change that isn't there.)
- **No CHANGELOG entry**, on a PR that changes a served tool schema and rejects inputs that previously reached the API. That's the most user-visible class of change this repo ships.

Also worth a deliberate decision either way: `get_log_attributes_for_pipeline` is the *recovery* tool — the one a model reaches for after `get_logs` rejects it. Giving it identical bare-field strictness risks a loop where both tools reject and the model has nowhere left to go.

Each fix is small and local. Not approving yet, but the plumbing and the message quality underneath are solid.

internal/telemetry/logs/query_sanitize_test.go
```diff
@@ -758,3 +759,466 @@ func TestGetLogsHandlerRejectsUnsupportedDottedRefsBeforeAPICall(t *testing.T) {
 		t.Fatalf("expected no API requests for invalid field refs, got %d", requestCount)
 	}
 }
+
```
faiz9068, line 762: Structural note on where these tests point: `sanitizeLogJSONQuery` is now called only from this file (28 call sites) — nothing in production reaches it directly anymore, everything goes through `prepareLogJSONQuery`. So every sanitize test here bypasses `validateLogJSONQuery` and the `$and`-wrap.

That's not hypothetical — it's the mechanism that let two of the Criticals above through. `TestUnknownKeyTipsAreStageSpecific` "proves" a tip the served tool can never emit, because it enters at a layer real requests skip. Worth collapsing `sanitizeLogJSONQuery` into `prepareLogJSONQuery` and repointing these tests at the real entry point, accepting the churn — otherwise ~500 lines of tests keep validating a path no request takes.

internal/prompts/prompts_test.go
```diff
@@ -45,6 +53,7 @@ func TestGetTracesDescriptionCriticalRules(t *testing.T) {
 		{"groupby", "must document groupby key name"},
 		{"resources['last9.tenant']", "must document tenant scoping"},
 		{"last9://reference/tracejson", "must point at the tracejson resource"},
+		{"60", "must document default lookback of 60 minutes"},
```
faiz9068: Minor: `"60"` as a bare substring will match `600`, `1600`, or any URL containing 60 — this check will essentially never fail. Assert on `default **60**` (the actual string the description now carries) so it can catch a regression.

dump_test.go
```diff
@@ -61,8 +61,14 @@ func TestDumpTools(t *testing.T) {
 		t.Fatal("get_logs description still contains unsubstituted {{labels}} placeholder")
 	}
 	for _, whale := range []string{"get_logs", "get_traces", "get_service_logs"} {
-		if len(out.Tools[byName[whale]].Description) > 2000 {
-			t.Fatalf("%s served description length %d exceeds 2000-char tripwire", whale, len(out.Tools[byName[whale]].Description))
+		whaleBudgets := map[string]int{
+			"get_logs":         4500,
```
faiz9068: Minor: 4500 is a lot of slack — the served `get_logs` description is 2471 bytes, so this leaves ~82% headroom on a tripwire whose entire purpose (per AGENTS.md's progressive-disclosure rule) is to catch regrowth. `get_traces` at 2200 for 1869 bytes is a sane ~18%. Suggest ~2600 for `get_logs`, matching the same ratio, in both this file and `prompts_test.go:88`.

Also `whaleBudgets` is declared inside the loop body, so it's rebuilt per iteration — moving it above the `for` reads better.

internal/prompts/descriptions/get_traces_base.md
```diff
@@ -12,7 +12,7 @@ Query traces with `tracejson_query` — a JSON **array of stages**. Each stage M
 
 **Scope:** tenant name → `resources['last9.tenant']`; deployment env → `resources['deployment.environment']`.
 
-**Time (tool args):** Prefer `lookback_minutes` (default **5**). Absolute → `start_time_iso`+`end_time_iso` (RFC3339). Never put the window as a `Timestamp` filter in the pipeline.
+**Time (tool args):** Prefer `lookback_minutes` (default **60**). Absolute → `start_time_iso`+`end_time_iso` (RFC3339). Never put the window as a `Timestamp` filter in the pipeline.
```
faiz9068, line 15: This fix is correct — `utils.DefaultLookbackMinutes = 60` (`internal/utils/utils.go:35`), so the old "5" was simply wrong. But `references/tracejson.md` wasn't updated alongside it: lines 38 and 42 still say `lookback_minutes: 5`, and line 447 says "CRITICAL: When no time is specified, MUST use lookback_minutes: 5 (NOT 60!)". Both files are served to the model and the manual is the more emphatic of the two, so as it stands this change makes the contradiction worse rather than resolving it.

internal/prompts/descriptions/get_logs_base.md
```diff
@@ -1,21 +1,28 @@
-Query logs with `logjson_query` — JSON **array of stages**. Each stage `"type"`: `filter`|`parse`|`aggregate`|`window_aggregate`. No `"stage"`/`"conditions"`.
+`logjson_query` is a JSON **array of stages** — NOT SQL, NOT a query string. Each stage sets `"type"`: `filter`|`parse`|`aggregate`|`window_aggregate`. No `"stage"`/`"conditions"`.
 
-**Filter:** `{"type":"filter","query":{"$and":[{"$eq":["SeverityText","ERROR"]}]}}` — `$eq|$neq|$contains|$gt|$gte|$lt|$lte|$regex` on `[field, value]` strings. Always `$and`-wrap.
+**Order rule:** filter first → parse (before using extracted attrs) → aggregate or window_aggregate only for counts/trends. Never use SpanKind/StatusCode/Duration/SpanName as log filters.
```
faiz9068, line 3: The description gains an "Order rule", but `validateLogJSONQuery` doesn't enforce any stage ordering — an `aggregate`-first pipeline passes, and so does a `filter` on a Body-derived attribute placed before the `parse` that creates it. That's a defensible MVP scope call given `logjson.md:33-36` calls ordering "CRITICAL", but the PR description reads as though it was implemented, so worth stating explicitly as deferred.

internal/telemetry/logs/input_schema.go
prathamesh-sonpatki: Fixed in 8d53762: dropped stage-level `additionalProperties:false` and loosened parse/WA schema `required` to `["type"]` so `window_minutes`/`format`/`aggregates` reach `validateLogJSONQuery` tips (live smoke confirms). Root `additionalProperties:false` kept.

internal/telemetry/logs/input_schema.go
prathamesh-sonpatki: Fixed in 8d53762: `oneOf` → `anyOf` for OpenAI structured-output compatibility.

internal/telemetry/logs/input_schema.go
prathamesh-sonpatki: Fixed in 8d53762: parse `field` stays schema-optional; `prepareLogJSONQuery` defaults missing/blank field to `"Body"` (aligned with logjson.md + upstream).

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed in 8d53762: dropped TraceId/SpanId/ParentSpanId from `traceOnlyLogFields` and added them to `isCanonicalLogFieldRef`. SpanKind/StatusCode/Duration/SpanName still rejected.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed in 8d53762: no longer require parse `field` — `prepareLogJSONQuery` defaults blank/missing to `"Body"`.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed in 8d53762: json/logfmt label keys validated with `safeBodyKeyPattern` (allows `http.status_code`); `validNamedCaptureName` kept for regexp captures only. Added hint→prepare round-trip test.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed in 8d53762: json/logfmt label keys validated with `safeBodyKeyPattern` (allows `http.status_code`); `validNamedCaptureName` kept for regexp captures only. Added hint→prepare round-trip test.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed in 8d53762: removed `windowLen`; gofmt applied.

internal/telemetry/logs/schema_test.go
prathamesh-sonpatki: Fixed in 8d53762: positive schema docs per stage type + tip-reachable cases (`window_minutes` without `window`, `format` without `parser`, aggregates-on-WA) must pass `resolved.Validate`.

internal/telemetry/logs/query_sanitize_test.go
prathamesh-sonpatki: Acknowledged in 8d53762: tip/unknown-key coverage goes through `prepareLogJSONQuery`; added TraceId/dotted-label/hint round-trip prepare tests. Full sanitize→prepare test collapse left as follow-up to keep this review commit focused.

internal/prompts/prompts_test.go
prathamesh-sonpatki: Fixed in 8d53762: assert `default **60**`; get_logs whale budget tightened to 2600.

dump_test.go
prathamesh-sonpatki: Fixed in 8d53762: get_logs budget 4500→2600; `whaleBudgets` moved above the loop.

internal/prompts/descriptions/get_traces_base.md
prathamesh-sonpatki: Fixed in 8d53762: `tracejson.md` lookback defaults updated 5→60; removed the emphatic NOT-60 contradiction.

internal/prompts/descriptions/get_logs_base.md
prathamesh-sonpatki: Acknowledged in 8d53762: stage order remains docs-only (MVP). Noted as deferred in the PR body; added a code comment on `validateLogJSONQuery`.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Correction: this thread is the typed-error surface. Fixed in 8d53762 — sanitize-layer rejections now return `*LogPipelineValidationError` with Category/Path (same taxonomy as validate).

internal/telemetry/logs/input_schema.go
```diff
@@ -0,0 +1,158 @@
...
+			},
+		},
+	}
+}
+
+func logFilterStageSchema() map[string]interface{} {
+	return map[string]interface{}{
+		"type":        "object",
+		"description": "Filter stage: narrows the dataset by condition. Always wrap in $and.",
+		"required":    []string{"type", "query"},
```
faiz9068: **Important — the loosening stopped half way, so three documented mistakes are still opaque.**

`parse` and `window_aggregate` now require only `type`, which is what made their tips reachable. But `filter` still requires `query` (this line), `aggregate` still requires `aggregates` (line 116), and the aggregate item still requires `function`+`as` (line 100). Because every branch pins `type` via an enum, a stage with the right `type` and a wrong key can only match its own branch — so it fails all four and the model gets the same opaque error you just eliminated for the other two.

I resolved the schema and ran these through it. In each case the validator already has a good message waiting that the model never sees:

| input | schema | validator's (unreachable) message |
|---|---|---|
| `{"type":"filter","conditions":{...}}` | opaque `anyOf` | `unknown key "conditions" on filter stage at logjson_query[0] — allowed keys: query, type` |
| `{"type":"aggregate","aggs":[...]}` | opaque `anyOf` | `unknown key "aggs" on aggregate stage at logjson_query[0] — allowed keys: aggregates, groupby, type` |
| `{"stage":"filter","query":{...}}` | opaque `anyOf` | `missing or non-string "type" at logjson_query[0] — each stage must set type to filter\|parse\|aggregate\|window_aggregate` |

The first and third are exactly the mistakes `get_logs_base.md` calls out by name ("No `\"stage\"`/`\"conditions\"`"), so these are the ones worth having a tip for.

Suggest dropping `filter` to `required: ["type"]` and `aggregate` to `required: ["type"]` for the same reason the other two were dropped — the validator already enforces both and says why.

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,444 @@
...
+				return err
+			}
+		case "window_aggregate":
+			if err := validateWindowAggregateStage(stage, stagePath); err != nil {
+				return err
+			}
+			if err := validateGroupByForTraceFields(stage, stagePath); err != nil {
+				return err
+			}
+		case "aggregate":
```
faiz9068, line 220: **Important — aggregate item keys are unvalidated, so `alias` is silently forwarded upstream.**

The `aggregate` case only runs `validateGroupByForTraceFields`; there is no `validateAggregateStage`, so nothing checks the keys inside an `aggregates[]` item. Today that's masked by `input_schema.go:100` requiring `function`+`as` — but masked opaquely, which is the problem, and it means the gap is invisible until the schema is loosened.

Executed, so this is measured rather than inferred. For `{"type":"aggregate","aggregates":[{"function":{"$count":[]},"alias":"c"}]}`:

- schema: `anyOf: did not validate against any of [...]` — opaque, no mention of `alias`
- `prepareLogJSONQuery`: **ACCEPTED, no error**

So if you loosen line 100 (or if a client tolerates the schema), `alias` goes straight to the API. `get_logs_base.md` and the aggregate stage description both say "Use `as`, NOT `alias`" — it's a known model mistake with no server-side rule behind it.

Suggest a small `validateAggregateStage` that walks `aggregates[]` and rejects unknown item keys with the same tip style as the stage-level check (`unknown key "alias" in aggregates[0] at … — allowed keys: as, function; use "as", not "alias"`). That closes the gap and lets line 100 be loosened alongside the sibling in my other comment.

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,444 @@
...
+	if pathPrefix == "" {
+		pathPrefix = "logjson_query"
+	}
+
+	// Validate first (fail-closed), then sanitize field refs, then $and-wrap.
+	if err := validateLogJSONQuery(stages, pathPrefix); err != nil {
+		return nil, err
+	}
+
+	// Default missing or blank parse field to "Body" after validation so the
```
faiz9068: Minor: this loop writes into the caller's stage maps rather than the rebuilt copies. Everything else in `prepareLogJSONQuery` is careful about that — `sanitizeLogJSONQueryPrefixed` rebuilds every stage map and every nested condition map specifically so the `$and`-wrap never touches `args`. I verified the difference: after `prepareLogJSONQuery(stages, …)` on `{"type":"parse","parser":"json"}`, the caller's own map is `map[string]interface{}{"field":"Body", "parser":"json", "type":"parse"}` — mutated.

Harmless today, since `args.LogjsonQuery` is discarded after the call. But it quietly breaks the no-aliasing property, and the next person who reads `args` after this point (retry logic, logging the original request, a caching layer) gets the normalized value instead of what the model sent. Cheap fix: do the defaulting inside the sanitize walk where the copy already exists, or shallow-copy the stage here before writing.

faiz9068 (commented):
Re-reviewed the delta `5555f5e..8d53762`. **All four Criticals are fixed, and I verified each by execution rather than by reading the diff** — I resolved the served schema the way the SDK does and ran the exact inputs through `prepareLogJSONQuery` in a throwaway worktree. Head CI is green (Run Tests ×2, CodeQL, Analyze go/js-ts/actions all success); locally `go build`, `go vet` and the full suite pass, and `gofmt -l` no longer flags either new file.

What I confirmed, with the actual output:

- **Tips are now reachable.** `window_minutes` and `aggregates` on window_aggregate, and `format` on parse, all pass schema validation and land on the real message — e.g. `unknown key "window_minutes" on window_aggregate stage at logjson_query[0] — allowed keys: as, function, groupby, type, window; common mistake: window_aggregate uses "function"+"as"+"window" not "aggregates" or "window_minutes"`. That was the whole point of the finding.
- **`TraceId` / `SpanId` / `ParentSpanId`** are accepted in both filter and `groupby`, and you moved *both* lists together — adding them to `isCanonicalLogFieldRef` is what avoids the trap where the bare-field rule rejects them instead, producing the same symptom from a different rule. Good catch on that.
- **`parse` without `field`** is accepted and goes upstream as `"field":"Body"`. The two `logjson.md` examples that were rejected now pass.
- **Dotted `labels` keys** work, and the boundaries are right: `http.status_code`, `user.id`, `a@b`, `a/b`, `_ok` accepted; `$merchant`, `has space`, `quo"te`, `back\slash`, `-leading` still rejected. `validNamedCaptureName` is correctly still used for regexp named captures.

Also resolved: `anyOf`, the typed error wired through the whole sanitize layer, `windowLen` deleted, `tracejson.md` corrected (no `lookback_minutes: 5` left anywhere), CHANGELOG added at the top of `[Unreleased]`, budgets cut to 2600 with the map hoisted out of the loop, `default **60**` replacing the bare `"60"` substring, per-stage positive satisfiability tests, the stage-ordering deferral documented in code, and `TestPrepareLogJSONQueryRoundTripBodyDerivedHint` — the discovery→query round-trip. All four tests the commit implies do exist; I grepped for each.

Two Important items remain, both in the area the fix touched, plus one minor. Details inline. The short version: the loosening was applied to `parse` and `window_aggregate` but not to `filter` or `aggregate`, so three *documented* mistakes still produce the opaque `anyOf` error even though the validator has a good message ready for each — and one of them turns out to expose a real validator gap (`alias` is silently forwarded upstream).

One thing I chased and am explicitly **not** raising against this PR: canonical `{"$count":[]}` reaches the wire as `{"$count":null}`, because `sanitizeLogFunction` does `append([]interface{}(nil), args...)` and appending zero elements leaves the slice nil. I confirmed that line is present at the merge base `c3f7734`, so it predates this PR, and your live/eval runs show upstream tolerates it. Flagging only so it's on the record — worth a separate look sometime, not here.

Not approving (that's not mine to give here), but the Criticals are genuinely closed and the remaining two are small and local.

internal/telemetry/logs/input_schema.go
prathamesh-sonpatki: Fixed: added `validateAggregateStage` — requires `aggregates`, rejects unknown item keys with `alias`→`as` tip, requires `function`+`as`.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed: added `validateAggregateStage` — requires `aggregates`, rejects unknown item keys with `alias`→`as` tip, requires `function`+`as`.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed: added `validateAggregateStage` — requires `aggregates`, rejects unknown item keys with `alias`→`as` tip, requires `function`+`as`.

internal/telemetry/logs/input_schema.go
prathamesh-sonpatki: Correction: this thread is the schema loosening. Fixed — filter/aggregate stage `required` is now only `["type"]`; aggregate item no longer schema-requires `function`/`as`. `conditions`/`aggs`/`alias` pass schema and hit validate tips.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Correction: this thread is the caller-map mutation. Fixed — parse `field`→`Body` defaulting moved into `sanitizeLogJSONQueryPrefixed` on rebuilt stage copies; caller maps are no longer mutated.

internal/telemetry/logs/input_schema.go
```diff
@@ -0,0 +1,163 @@
...
+		"additionalProperties": false,
+	}
+}
+
+func logjsonQuerySchema() map[string]interface{} {
+	return map[string]interface{}{
+		"type":        "array",
+		"description": "JSON pipeline query for logs. An ordered list of stages: filter → parse → aggregate/window_aggregate. NOT SQL, NOT a query string. Each stage is an object with a 'type' field.",
+		"items": map[string]interface{}{
+			"anyOf": []interface{}{
```
faiz9068, line 55: **The loosening covers unknown keys, but not wrong types or enums — so a set of mistakes still gets the opaque error.**

Every branch still constrains value types (`query: object`, `function: object`, `window: array`, `groupby: object`) and enums (`parser`, `type`). Anything violating one fails all four branches at the SDK layer, before the handler runs. Resolved the served schema and ran each case — 7 of 7 opaque, and 5 have a good validator message the model never sees:

| input | unreachable message |
|---|---|
| `{"type":"filter","query":"SELECT * FROM logs"}` | *(accepted — see the other comment)* |
| `{"type":"sort",…}` | `unknown stage type "sort" … valid types are filter, parse, aggregate, window_aggregate` |
| stage with no `type` | `missing or non-string "type" …` |
| `"parser":"grok"` | `invalid parser "grok" … must be one of: json, logfmt, regexp` |
| `"function":"$count"` | `"function" must be an object like {"$count":[]}, not a string` |
| `"window":"1m"` | `"window" must be an array of exactly 2 elements …` |

The NOT-SQL one matters most — it's the first line of `get_logs_base.md`, and the model gets `anyOf: did not validate against any of [<anonymous schema> ×4]`.

Fix in the same spirit as the rest of the file — a permissive catch-all branch, so the four descriptive branches still ship as guidance in `tools/list` but rejection is owned entirely by `validateLogJSONQuery`:

```go
"anyOf": []interface{}{
    logFilterStageSchema(), logParseStageSchema(),
    logAggregateStageSchema(), logWindowAggregateStageSchema(),
    map[string]interface{}{"type": "object"},
},
```

Worth adding these six cases to `schema_test.go` asserting the surfaced error contains the tip and not `anonymous schema` — the current `tipReachable` table is all extra-key cases, which is the one class that already works.

faiz9068 (commented):
Took another pass over the whole PR rather than just the delta, and it turned up a class I'd missed in earlier rounds — my earlier checks only covered wrong-*key* mistakes, which this PR does fix. Wrong *type* and wrong *enum* are still opaque, including the NOT-SQL case. Two findings inline, both verified by execution.

Build, vet and the full suite pass at this head; gofmt clean on the PR's files.

internal/telemetry/logs/query_validate.go
```diff
@@ -0,0 +1,550 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// validateFilterStage checks that a filter stage has a non-nil "query" key and
+// then delegates trace-field checks to validateFilterStageForTraceFields.
+func validateFilterStage(stage map[string]interface{}, stagePath string) error {
```
faiz9068, line 443: **Two shapes are fail-open, and on the attributes path they reach the API.**

`validateFilterStage` only checks `query != nil`, and `validateGroupByForTraceFields` (line 306) returns `nil` when `groupby` isn't a map. Both are accepted and forwarded:

```
filter query is SQL string -> ACCEPTED, sent upstream: [{"query":"SELECT * FROM logs","type":"filter"}]
groupby is an array        -> ACCEPTED, sent upstream: [{…,"groupby":["ServiceName"],…}]
groupby is a string        -> ACCEPTED, sent upstream: [{…,"groupby":"ServiceName",…}]
```

On `get_logs` the hand-written schema masks this (opaquely — the other comment). On `get_log_attributes_for_pipeline` nothing masks it: `tools.go:138` registers a hand-written schema for `get_logs` only, and that tool's `pipeline` items are `additionalProperties: true`. So SQL-in-a-filter goes straight upstream and comes back as an opaque 400 — which is the case fail-closed validation is meant to prevent.

Fix: require `query` to be a `map[string]interface{}` with a "NOT SQL — `query` is a condition object" tip, and reject a non-map `groupby` instead of returning nil.

dump_test.go
```diff
@@ -105,9 +105,15 @@ func TestDumpTools(t *testing.T) {
 	if strings.Contains(out.Tools[byName["get_logs"]].Description, "{{labels}}") {
 		t.Fatal("get_logs description still contains unsubstituted {{labels}} placeholder")
 	}
+	whaleBudgets := map[string]int{
+		"get_logs":         2600,
```
faiz9068: Minor: this budget measures `description` only, but the hand-written InputSchema is now the larger half. Measured from `dump-tools` at this head: `get_logs` is desc 2423 + schema 3860 = **6283 B** served, so the part that grew most is unguarded. Suggest measuring `len(description) + len(json(inputSchema))`.

Also `get_traces` (2200) and `get_service_logs` (2000) were raised here though neither description changed — worth putting those back.

internal/telemetry/logs/input_schema.go
prathamesh-sonpatki: Fixed: added permissive catch-all `{"type":"object"}` anyOf branch so wrong types/enums (SQL query, sort, grok, string function/window) pass schema and hit validate tips. Live smoke confirmed — no opaque anonymous-schema errors. Tests cover the six cases.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: Fixed: `validateFilterStage` requires `query` to be a condition object (NOT SQL tip); `validateGroupByForTraceFields` rejects non-map groupby instead of returning nil.

dump_test.go
prathamesh-sonpatki: Fixed: `get_logs` dump budget now measures `len(description)+len(json(inputSchema))` (7000 ceiling). Restored `get_traces`/`get_service_logs` description budgets to 2000.

faiz9068 (approved):

PR #213, opened 2026-08-14
fix(mcp): compile last9/api fetches and structured log search

prathamesh-sonpatki:
## Summary
- Align served whale descriptions with last9/api (`window_aggregate` is `function`/`as`/`window`; traces lookback 60) so description-only agents stop sending `aggregates`+`window_minutes` (ENG-1730).
- Honest failures: PromQL/logs 400s relay the sanitized API body; 5xx bodies are drained; any failed log/trace chunk is `isError` instead of OK `partial_result`.
- `get_traces` fail-closes unknown stages and illegal `window_aggregate` keys before upstream. `get_service_logs` compiles HTTP status class/code and named-attribute filters server-side. On-call / `get_exceptions` route status-class log search to `get_service_logs`.

Waterfall `evidence.partial` and deviations `partial_errors` are unchanged. Logs fail-closed stage allowlist from ENG-1729 is **not** in this PR (ships in #211).

## Test plan
- [x] `go test ./...`
- [x] `go test -run 'TestDumpTools|TestDumpToolsInvestigate|TestOnCallRunbook' .`
- [x] Live MCP against this branch: illegal trace `window_aggregate` local error; PromQL 400 includes parse body; `get_service_logs` invalid attribute syntax local error; HTTP 5xx discovers or asks for `http_status_field`
- [ ] Restart Cursor `last9_local` and try HTTP 5xx / bad PromQL in a fresh chat
- [ ] Follow-up: flip `log_tool_selection` evals from `get_logs` to `get_service_logs` for HTTP status (current fixtures would fail on purpose)

### DIFF
--- a/internal/apm/apm.go
+++ b/internal/apm/apm.go
@@ -274,7 +274,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			data, err := io.ReadAll(httpResp.Body)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to read response body: %w", err)
@@ -297,7 +300,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			data, err := io.ReadAll(httpResp.Body)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to read response body: %w", err)
@@ -320,7 +326,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			data, err := io.ReadAll(httpResp.Body)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to read response body: %w", err)
@@ -343,7 +352,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			// read response body to byte array
 			data, err := io.ReadAll(httpResp.Body)
 			if err != nil {
@@ -367,7 +379,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			data, err := io.ReadAll(httpResp.Body)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to read response body: %w", err)
@@ -389,7 +404,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			data, err := io.ReadAll(httpResp.Body)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to read response body: %w", err)
@@ -411,7 +429,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			var topErrResp apiPromInstantResp
 			if err := json.NewDecoder(httpResp.Body).Decode(&topErrResp); err == nil {
 				details.TopOperations.ByResponseTime = make([]map[string]float64, 0)
@@ -450,7 +471,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			var topErrResp apiPromInstantResp
 			if err := json.NewDecoder(httpResp.Body).Decode(&topErrResp); err == nil {
 				details.TopOperations.ByErrorRate = make([]map[string]int64, 0)
@@ -488,7 +512,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 			return nil, nil, err
 		}
 		defer httpResp.Body.Close()
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
+			return nil, nil, promErr(httpResp, "service performance details")
+		}
+		{
 			var topErrResp apiPromInstantResp
 			if err := json.NewDecoder(httpResp.Body).Decode(&topErrResp); err == nil {
 				details.TopErrors = make([]map[string]int64, 0)
@@ -561,7 +588,7 @@ func NewServiceOperationsSummaryHandler(client *http.Client, cfg models.Config) 
 		}
 		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to get service operations summary: %s", httpResp.Status)
+			return nil, nil, promErr(httpResp, "service operations summary")
 		}
 		var promResp apiPromInstantResp
 		if err := json.NewDecoder(httpResp.Body).Decode(&promResp); err != nil {
@@ -579,7 +606,7 @@ func NewServiceOperationsSummaryHandler(client *http.Client, cfg models.Config) 
 		}
 		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to get service operations summary: %s", httpResp.Status)
+			return nil, nil, promErr(httpResp, "service operations summary")
 		}
 		var respTimeRaw apiPromInstantResp
 		if err := json.NewDecoder(httpResp.Body).Decode(&respTimeRaw); err != nil {
@@ -598,7 +625,7 @@ func NewServiceOperationsSummaryHandler(client *http.Client, cfg models.Config) 
 		}
 		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to get service operations summary: %s", httpResp.Status)
+			return nil, nil, promErr(httpResp, "service operations summary")
 		}
 		var errorRateRaw apiPromInstantResp
 		if err := json.NewDecoder(httpResp.Body).Decode(&errorRateRaw); err != nil {
@@ -616,7 +643,7 @@ func NewServiceOperationsSummaryHandler(client *http.Client, cfg models.Config) 
 		}
 		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to get service operations summary: %s", httpResp.Status)
+			return nil, nil, promErr(httpResp, "service operations summary")
 		}
 		var dbThroughputRaw apiPromInstantResp
 		if err := json.NewDecoder(httpResp.Body).Decode(&dbThroughputRaw); err != nil {
@@ -634,7 +661,7 @@ func NewServiceOperationsSummaryHandler(client *http.Client, cfg models.Config) 
 		}
 		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to get service operations summary: %s", httpResp.Status)
+			return nil, nil, promErr(httpResp, "service operations summary")
 		}
 		var dbRespTimeRaw apiPromInstantResp
 		if err := json.NewDecoder(httpResp.Body).Decode(&dbRespTimeRaw); err != nil {
@@ -668,7 +695,7 @@ func NewServiceOperationsSummaryHandler(client *http.Client, cfg models.Config) 
 		}
 		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to get service operations summary: %s", httpResp.Status)
+			return nil, nil, promErr(httpResp, "service operations summary")
 		}
 		var dbErrorRateRaw apiPromInstantResp
 		if err := json.NewDecoder(httpResp.Body).Decode(&dbErrorRateRaw); err != nil {
@@ -686,7 +713,7 @@ func NewServiceOperationsSummaryHandler(client *http.Client, cfg models.Config) 
 		}
 		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to get service operations summary: %s", httpResp.Status)
+			return nil, nil, promErr(httpResp, "service operations summary")
 		}
 		var httpThroughputRaw apiPromInstantResp
 		if err := json.NewDecoder(httpResp.Body).Decode(&httpThroughputRaw); err != nil {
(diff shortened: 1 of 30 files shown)


internal/telemetry/logs/logs.go
```diff
@@ -357,8 +334,7 @@ func executeLogJSONQuery(ctx context.Context, client *http.Client, cfg models.Co
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		body, _ := io.ReadAll(resp.Body)
-		return nil, fmt.Errorf("logs API request failed with status %d: %s", resp.StatusCode, string(body))
+		return nil, utils.NewUpstreamHTTPError(resp, "logs query")
```
faiz9068: **Blocker: this line is unreachable for a 400, so logs 400s are still relayed unsanitized.**

`utils.MakeLogsJSONQueryAPI` intercepts 400 itself and returns `fmt.Errorf("logs API request failed with status %d: %s\n\n%s", ..., string(respBody), pipelineSchemaHint)` — raw `io.ReadAll`, no redaction, no size bound. Only 422 and non-400 statuses reach `NewUpstreamHTTPError` here.

So for `get_logs` / `get_service_logs` (and the same call in `internal/apm/databases.go`), a 400 body reaches the model verbatim: URLs and bearer tokens included, unbounded length, and it also lands in the `slog.Error` line. That's the opposite of what the PR claims for the logs paths.

`http_error_test.go`'s "400 includes parse body" case passes via the *old* path, so it reads as green while proving nothing about the new helper.

Suggested fix: drop the 400 special case in `MakeLogsJSONQueryAPI` and return the `*http.Response` so the caller decides (traces already does this), then give `NewUpstreamHTTPError` an optional hint arg so `pipelineSchemaHint` still rides along. And make the test assert redaction actually ran — a body with a URL should come back `[redacted-url]`, an over-long one `… (truncated)`.

internal/telemetry/logs/service_logs_filters.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	if len(matches) > 1 {
+		names := make([]string, 0, len(matches))
+		for _, attr := range matches {
+			names = append(names, attr.FilterField)
+		}
+		return "", nil, fmt.Errorf("multiple HTTP status fields found (%s); pass http_status_field with the filter_field to use", strings.Join(names, ", "))
+	}
+
+	return matches[0].FilterField, parseStagesFromAttributeHint(matches[0]), nil
```
faiz9068, line 163: A single *wrong* match is indistinguishable from a correct one, and the agent can't see it.

`isHTTPStatusLikeAttribute` matches any `status_code` substring, so a service whose only match is e.g. `attributes['job_status_code']` or `attributes['task_status_code']` yields exactly one hit. Discovery "succeeds", `$regex ^5` gets applied to a non-HTTP field, and the tool returns a confident wrong answer. The ambiguity guard below can't help — there's no ambiguity.

`ServiceLogsResponse` reports neither the resolved field nor the compiled condition, so nothing downstream can sanity-check it. Echoing the resolved field back (`"http_status_field": "attributes['status_code']"`) is cheap and makes discovery auditable in transcripts.

internal/telemetry/logs/service_logs_filters.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func applyServiceLogsStructuredFilters(query []map[string]interface{}, extra []map[string]interface{}, parseStages []map[string]interface{}) []map[string]interface{} {
+	if len(extra) == 0 && len(parseStages) == 0 {
+		return query
+	}
+
+	if len(parseStages) == 0 {
+		if len(query) == 0 {
+			return query
```
faiz9068: These three bailouts (`208`, `213`, `218`) return the query with the caller's filters **silently dropped** — `get_service_logs` would return every log for the service while the caller believes it asked for 5xx. A false-clean window, which is the bug class this PR removes elsewhere.

Unreachable today since `buildServiceLogsQuery` always emits `{type: filter, query: {$and: [...]}}`, so this is latent rather than live — but it's ~5 lines to return an error instead of degrading quietly. (Same critique applies to the pre-existing `addServiceLogsEnvFilter` this was patterned on.)

Minor, same function: the no-parse path mutates `query[0]` in place while the parse path copies. No live bug — both call sites reassign — but worth picking one discipline so a future caller reusing the input slice isn't surprised.

internal/telemetry/logs/service_logs_filters.go
```diff
@@ -0,0 +1,245 @@
...
+	return "", fmt.Errorf("invalid log field reference %q at %s: use ServiceName, attributes['field'], or resources['field']", field, path)
+}
+
+func httpStatusCondition(field, class, code string) map[string]interface{} {
+	if httpStatusCodePattern.MatchString(code) {
+		return map[string]interface{}{
+			"$eq": []interface{}{field, code},
+		}
+	}
+	digit := httpStatusClassPattern.FindStringSubmatch(class)[1]
```
faiz9068: `FindStringSubmatch(class)[1]` indexes without a nil check. Unreachable today (both args are validated above), but it's a panic one refactor away.

internal/telemetry/logs/service_logs.go
```diff
@@ -330,37 +340,27 @@ func fetchServiceLogs(ctx context.Context, client *http.Client, cfg models.Confi
...
 			service,
 			len(logs),
 			startMs,
 			endMs,
-			partialErr != nil,
 		)
 	}
 
-	response := &ServiceLogsResponse{
+	return &ServiceLogsResponse{
```
faiz9068, line 360: `PartialResult` and `Warning` are still declared on `ServiceLogsResponse` (:28-29) even though nothing writes them now — `omitempty` means no wire impact, but a future reader will assume the partial contract still exists. Worth deleting with the rest.

Also: the four inverted tests still carry their old names (`TestFetchServiceLogsReturnsPartialResultsAfterLaterChunkError` and friends) and now assert the opposite of what they say.

internal/apm/apm.go
```diff
@@ -1702,10 +1738,10 @@ func NewPromqlLabelValuesHandler(client *http.Client, cfg models.Config) func(co
 		if httpResp == nil {
 			return nil, nil, fmt.Errorf("received nil response from Prometheus")
 		}
+		defer httpResp.Body.Close()
 		if httpResp.StatusCode != http.StatusOK {
-			return nil, nil, fmt.Errorf("failed to execute Prometheus range query: %s", httpResp.Status)
+			return promToolError(httpResp, "Prometheus range query")
```
faiz9068: Op label copy-pasted from the range-query handler — this is `NewPromqlLabelValuesHandler`, so errors will be mislabeled "Prometheus range query". Same at :1786 in `NewPromqlLabelsHandler`.

internal/apm/apm.go
```diff
@@ -274,7 +274,10 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 		}
 		defer httpResp.Body.Close()
 
-		if httpResp.StatusCode == http.StatusOK {
+		if httpResp.StatusCode != http.StatusOK {
```
faiz9068, line 278: This makes `get_service_performance_details` all-or-nothing across nine sequential PromQL calls — previously a non-200 was skipped and the surviving sub-metrics were returned. `TestServicePerformanceDetailsAnyPromFailureIsError` locks it in, so it looks deliberate, but it isn't in the PR's stated scope: one flaky sub-query now returns nothing instead of throughput + latency + apdex. Intended? A per-section error field might serve callers better than total failure.

Cosmetic: the `if != OK { return }` followed by a bare `{ ... }` block, repeated nine times, reads like a merge artifact — the leftover braces can just be dedented.

internal/telemetry/traces/errors.go
```diff
@@ -181,21 +174,5 @@ func readLimitedResponseBody(body io.Reader, limit int64) string {
 }
 
 func sanitizeUpstreamBody(raw string) string {
```
faiz9068: Only `SanitizeUpstreamBody` was promoted — `drainResponseBody` and `readLimitedResponseBody` above are still duplicated verbatim from `internal/utils/upstream_http.go`. Export those two from `utils` and drop the local copies, otherwise the shared-helper consolidation is two-thirds done.

internal/prompts/descriptions/get_traces_base.md
```diff
@@ -1,4 +1,4 @@
-Query traces with `tracejson_query` — a JSON **array of stages**. Each stage MUST set `"type"` to `filter`|`parse`|`aggregate`|`window_aggregate`. Do **not** use `"stage"` or `"conditions"`. For an exact `trace_id`, prefer `get_service_traces`.
+Query traces with `tracejson_query` — a JSON **array of stages**. Each stage MUST set `"type"` to `filter`|`parse`|`aggregate`|`window_aggregate`. Do **not** use `"stage"` or `"conditions"`. For an exact `trace_id` or recent traces of one service, prefer `get_service_traces`.
```
faiz9068, line 1: This description still never teaches the `window_aggregate` *shape*: it documents `aggregate` as `aggregates` + `groupby`, names `window_aggregate` as a legal `type`, and stops there. A description-only agent's most likely guess is to reuse `aggregates` on a `window_aggregate` stage — exactly the failure this PR fixes for `get_logs`.

The new validator now returns a good corrective error, so it costs a self-correction round rather than a bad query, but one line here (plus the matching `dump_test.go` assertion for `get_traces`, which currently pins `function`/`as`/`window` for `get_logs` only) closes the loop.

faiz9068 (commented):
Reviewed the fetch-contract changes. The fail-closed work is solid and the zero-upstream-request tests are the right kind of proof.

One blocker: the sanitized-400 relay never runs on the logs paths (inline on `logs.go`). Details inline; the rest are should-fix / minor.

Two things not anchorable to a line:

1. **No CHANGELOG entry.** Dropping `partial_result` / `_last9_mcp` from `get_logs`/`get_traces`/`get_service_logs`, and turning a chunk hole into an error, is a breaking output-contract change. Needs a `### Changed` **Breaking:** bullet at the top of the section, plus a line for the new `get_service_logs` args.
2. **Eval fixtures are now knowingly red.** The descriptions were inverted to route HTTP-status search to `get_service_logs`, so `log_tool_selection` fixtures asserting `get_logs` will fail. Worth flipping here or filing the follow-up before merge so the suite isn't left failing.

Verified locally: `go build ./...` and `go test ./...` pass at head.

prathamesh-sonpatki:
Addressed the full review in a0b16f8.

CHANGELOG Unreleased now records the breaking chunk-hole / `partial_result` drop, the new `get_service_logs` status/attribute args, per-section `partial_errors` on `get_service_performance_details`, and sanitized log 400s.

HTTP-status `log_tool_selection` fixtures are flipped in last9/last9-mcp-evals#50 (`lts-001`/`002`/`006`/`011`/`013` → `get_service_logs`; gRPC / `hook_event` / `response_time` stay on `get_logs`).

internal/telemetry/logs/logs.go
prathamesh-sonpatki: Fixed in a0b16f8. `MakeLogsJSONQueryAPI` now always returns the `*http.Response`; `executeLogJSONQuery` maps non-OK through `NewUpstreamHTTPError` with `LogsPipelineSchemaHint`. Tests assert `[redacted-url]`, `… (truncated)`, and no leaked URL/SECRET.

internal/telemetry/logs/service_logs_filters.go
prathamesh-sonpatki: Fixed. Discovery now requires `http`+`status` together, or an exact base of `status_code`/`statuscode`/`status` (so `job_status_code` does not win). gRPC names stay excluded. The response echoes `http_status_field`.

internal/telemetry/logs/service_logs_filters.go
prathamesh-sonpatki: Fixed. Missing query/`$and` is now a tool error instead of a silent drop, and the filter path copies stages instead of mutating `query[0]`.

internal/telemetry/logs/service_logs_filters.go
prathamesh-sonpatki: Fixed. `httpStatusCondition` returns an error when `FindStringSubmatch` does not produce a class digit.

internal/telemetry/logs/service_logs.go
prathamesh-sonpatki: Removed `PartialResult`/`Warning`. Chunk-hole tests are renamed to `ReturnsErrorAfterLaterChunk…`.

internal/apm/apm.go
prathamesh-sonpatki: Fixed. Label-values errors say `Prometheus label values`; labels errors say `Prometheus labels`. Range query is unchanged.

internal/apm/apm.go
prathamesh-sonpatki: Fixed. Each of the nine PromQL HTTP non-OK responses appends a sanitized message to `partial_errors` and keeps surviving sections. Transport errors still fail the tool. Test now expects success JSON with the 400 body in `partial_errors`.

internal/telemetry/traces/errors.go
prathamesh-sonpatki: Exported `DrainResponseBody` / `ReadLimitedResponseBody` from `utils`; traces now calls those instead of local copies.

internal/prompts/descriptions/get_traces_base.md
prathamesh-sonpatki: Taught the `window_aggregate` shape (`function`/`as`/`window`) on `get_traces_base.md` (still under the 2000-char budget, no trailing newline). `dump_test.go` pins the same keys as `get_logs`.

faiz9068 (commented):
Re-reviewed `64dd730..a0b16f8` only. Thanks — 8 of 9 findings resolved, and a couple came back better than what I asked for.

Verified resolved:

- **The 400 blocker is genuinely fixed.** Interceptor removed from `MakeLogsJSONQueryAPI`, caller sanitizes, hint preserved via the variadic. I checked both non-test callers for the obvious same-class regression — a caller that only tested `err != nil` would now decode a 400 body as success — and `logs.go:330` and `databases.go:1005` both gate on `StatusCode != OK`. No regression. The new assertions (`[redacted-url]`, no `SECRET`, truncation marker) would catch a relapse, which the old test wouldn't have.
- `PartialResult`/`Warning` gone, replaced by the echoed `http_status_field` — that plus the per-section `partial_errors` on `get_service_performance_details` are both better than what I suggested.
- Silent filter-drops now error, and the in-place `query[0]` mutation is gone in favour of cloning every stage.
- `FindStringSubmatch` guarded; both op labels fixed; `drainResponseBody`/`readLimitedResponseBody` deduped into `utils`; `get_traces` description teaches the `window_aggregate` shape with a matching `dump_test` pin; CHANGELOG added.

`go build ./...` + `go test ./...` clean at `a0b16f8`, and all six checks green on that head.

**One finding is still live** — the status matcher fix closes the underscore spelling and opens the dotted one. Inline on `service_logs_filters.go`; that thread was resolved but the case isn't covered. Executed evidence and a one-line fix in the comment.

Smaller, non-blocking:

1. **CHANGELOG bullets went to the bottom of their sections.** Convention here puts a new entry at the **top** of its `###` section — so the #213 `Fixed` bullet above the #212 one, and the #213 `Changed` bullets above the `get_service_summary` entry.
2. The Breaking bullet lists `get_service_performance_details` `partial_errors` under "declared truncation … is unchanged", but this PR *adds* that field. Worth rewording so the entry doesn't read as pre-existing.
3. Two stale test names survive: `logs/chunking_test.go:491` `TestFetchServiceLogsReturnsEmptyPartialWhenAllSuccessChunksAreEmpty` and `traces/chunking_test.go:414` `TestGetTracesHandlerReturnsPartialResultAfterLaterChunkError`. Both assert errors now.
4. `fetchLogSeriesFieldNames` still inherits the caller ctx unbounded — only the body-sampling leg got `PerChunkHTTPTimeout`. On the new `get_service_logs` path a slow series call can still consume the budget before the real log query starts.
5. `scripts/eval-r10.sh` `log_tool_selection` fixtures are untouched, so still red by design — fine if the follow-up is filed, just shouldn't merge and forget.

No approval from me on this pass; happy to take another look once the matcher row lands.

internal/telemetry/logs/service_logs_filters.go
```diff
@@ -0,0 +1,267 @@
...
+		return false
+	}
+}
+
+func httpStatusAttributeBase(s string) string {
+	s = strings.ToLower(strings.TrimSpace(s))
+	s = strings.TrimPrefix(s, "attributes['")
+	s = strings.TrimPrefix(s, "resources['")
+	s = strings.TrimSuffix(s, "']")
+	if i := strings.LastIndexAny(s, "./"); i >= 0 {
```
faiz9068: Re-opening the status-matcher finding — this fix closes the spelling its own test names and opens the sibling.

Stripping to the last `.`/`/` segment means any dotted namespace collapses to a bare `status_code`. Ran the new predicate at this head:

```
status_code            httpStatusLike=true    base="status_code"
http.status_code       httpStatusLike=true    base="status_code"
job_status_code        httpStatusLike=false   base="job_status_code"   <- the tested case
job.status_code        httpStatusLike=true    base="status_code"      <- still wrong
db.status_code         httpStatusLike=true    base="status_code"
messaging.status_code  httpStatusLike=true    base="status_code"
queue.status           httpStatusLike=true    base="status"
grpc.status_code       httpStatusLike=false   base="status_code"      <- grpc guard, correct
```

Dotted is the OTel convention — it's how `http.status_code` itself is spelled — so `db.status_code` is not an exotic input. A service whose only match is that field still yields exactly one match, discovery still "succeeds", and `$regex ^5` still lands on a non-HTTP field. Same failure this finding was about, just via the more common spelling.

Fix is to delete the strip:

```go
	if i := strings.LastIndexAny(s, "./"); i >= 0 {
		s = s[i+1:]
	}
```

Without it, `job.status_code` → base `job.status_code` → false; `http.status_code` still matches on the earlier `http`+`status` check; bare `status_code` still hits the switch. Worth adding `job.status_code` and `db.status_code` as rows in `TestIsHTTPStatusLikeAttributeRejectsJobStatusCode`, since it currently only covers the underscore form.

prathamesh-sonpatki: Fixed in bab163a. Dropped the last-segment strip so `job.status_code` / `db.status_code` stay non-HTTP; `http.status_code` still matches on the `http`+`status` check and bare `status_code` still hits the switch. Added those dotted rows to the matcher test.

prathamesh-sonpatki:
Follow-up from the second pass is in bab163a.

- Matcher: no longer collapses dotted namespaces to a bare `status_code` suffix.
- CHANGELOG: #213 bullets moved to the top of Fixed/Changed; `partial_errors` on `get_service_performance_details` is worded as new, not as pre-existing truncation.
- Renamed the two leftover chunk tests that still said Partial.
- `fetchLogSeriesFieldNames` now uses `PerChunkHTTPTimeout` like the body-sampling leg.
- `log_tool_selection` fixture flip is last9/last9-mcp-evals#50 (so `scripts/eval-r10.sh` stays red until that merges).

faiz9068 (approved):

PR #214, opened 2026-08-14
feat(mcp): add continuous profiling tools (ENG-1067)

prathamesh-sonpatki:
## Summary
- Adds `get_profile_services`, `get_flamegraph`, `get_top_functions`, and `get_profile_summary` using the same PDE-718 `POST /profiles/api/v1/query_range/json` contract as the Profiling UI (last9#11392 / #11396 Tempo proxy).
- Registers a `profiles` toolset and includes it in `investigate`.
- Client-side flame tree + top-functions folding matches the dashboard ProfilesApis helpers (dataframe unwrap, CPU type pin, 0x1f frames).

## Test plan
- [x] `go test ./...`
- [x] `go run . dump-tools --toolsets=profiles` shows only the four tools
- [x] Live smoke on `app.last9.io` (`LAST9_LIVE_PROFILES=1`): services + flamegraph + top_functions + summary 200
- [ ] `go run . dump-tools --toolsets=logs` does not leak profiling tools

Linear: ENG-1067
Depends on: profiles Tempo proxy on app (last9#11396 / PDE-1117)

### DIFF
--- a/internal/telemetry/profiles/get_top_functions.go
+++ b/internal/telemetry/profiles/get_top_functions.go
@@ -49,19 +49,24 @@ func NewGetTopFunctionsHandler(client *http.Client, cfg models.Config) func(cont
 			rankLimit = DefaultTopFunctionsLimit
 		}
 
-		filters := filtersFromArgs(args.Service, args.Env, args.Cluster, args.Namespace, args.Runtime, args.ProfileType)
+		filters, err := filtersFromArgs(args.Service, args.Env, args.Cluster, args.Namespace, args.Runtime, args.ProfileType)
+		if err != nil {
+			return nil, nil, err
+		}
 		rows, err := runQueryRange(ctx, client, cfg, flamegraphPipeline(filters, DefaultFlamegraphRowLimit), start, end, DefaultFlamegraphRowLimit, args.Region)
 		if err != nil {
 			return utils.ToolErrorResult(fmt.Sprintf("failed to fetch top functions: %v", err)), nil, nil
 		}
 
-		functions := limitTopFunctions(FoldToTopFunctions(mapFlamegraphRows(rows)), rankLimit)
+		all := FoldToTopFunctions(mapFlamegraphRows(rows))
+		total := getProfileTotalSamples(all)
+		functions := limitTopFunctions(all, rankLimit)
 		result, err := jsonResult(map[string]any{
 			"service":        filters.Service,
 			"profile_type":   string(filters.ProfileType),
 			"start":          start.UTC().Format(time.RFC3339),
 			"end":            end.UTC().Format(time.RFC3339),
-			"total_samples":  getProfileTotalSamples(functions),
+			"total_samples":  total,
 			"function_count": len(functions),
 			"functions":      functions,
 		})
--- a/internal/telemetry/profiles/get_top_functions.go
+++ b/internal/telemetry/profiles/get_top_functions.go
@@ -1,0 +1,70 @@
+package profiles
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"last9-mcp/internal/models"
+	"last9-mcp/internal/utils"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// GetTopFunctionsArgs ranks hottest functions by self samples for a service.
+type GetTopFunctionsArgs struct {
+	Service         string `json:"service" jsonschema:"(Required) Service name (ServiceName) to profile"`
+	Env             string `json:"env,omitempty" jsonschema:"Filter by deployment.environment.name"`
+	Cluster         string `json:"cluster,omitempty" jsonschema:"Filter by k8s.cluster.name"`
+	Namespace       string `json:"namespace,omitempty" jsonschema:"Filter by k8s.namespace.name"`
+	Runtime         string `json:"runtime,omitempty" jsonschema:"Filter by telemetry.sdk.language"`
+	ProfileType     string `json:"profile_type,omitempty" jsonschema:"Profile type: cpu (default), alloc, or wall"`
+	Region          string `json:"region,omitempty" jsonschema:"Optional region override; defaults to the configured datasource region"`
+	Limit           int    `json:"limit,omitempty" jsonschema:"Max ranked functions to return after folding (default: 50)"`
+	LookbackMinutes int    `json:"lookback_minutes,omitempty" jsonschema:"Minutes to look back from now (default: 60, minimum: 1)"`
+	StartTimeISO    string `json:"start_time_iso,omitempty" jsonschema:"(Optional) Start time in RFC3339/ISO8601 format"`
+	EndTimeISO      string `json:"end_time_iso,omitempty" jsonschema:"(Optional) End time in RFC3339/ISO8601 format"`
+}
+
+// NewGetTopFunctionsHandler returns functions ranked by self sample share.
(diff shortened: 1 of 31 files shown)


internal/telemetry/profiles/get_top_functions.go
```diff
@@ -0,0 +1,75 @@
...
+		filters, err := filtersFromArgs(args.Service, args.Env, args.Cluster, args.Namespace, args.Runtime, args.ProfileType)
+		if err != nil {
+			return nil, nil, err
+		}
+		rows, err := runQueryRange(ctx, client, cfg, flamegraphPipeline(filters, DefaultFlamegraphRowLimit), start, end, DefaultFlamegraphRowLimit, args.Region)
+		if err != nil {
+			return utils.ToolErrorResult(fmt.Sprintf("failed to fetch top functions: %v", err)), nil, nil
+		}
+
+		all := FoldToTopFunctions(mapFlamegraphRows(rows))
```
nishantmodak: [P2] Disclose truncated input before reporting profile-wide rankings

Both this handler and `get_profile_summary` fold only the first 1,000 aggregated stacks. With 1,000 distinct 10-sample stacks followed by 1,001 9-sample stacks sharing the same leaf, they omit the actual hottest function (9,009 samples), report `total_samples=10000` instead of 19,009, and expose no indication that the ranking/percentages are partial. The description explicitly calls this the full-profile total, so an investigating agent can confidently recommend optimizing the wrong functions.

Propagate the row-cap/truncation signal to both outputs (as `get_flamegraph` already does), label totals/percentages and summary text as covering returned stacks when capped, and remove the unconditional full-profile guarantee. Alternatively compute the full aggregation before ranking. Increasing the cap alone does not fix the contract.

Reproduced twice on HEAD `f59f90b8ad2c0b352ba3ebf79ea4d474c0a58d02` using Go 1.27.1, locked `go.mod`/`go.sum`, synthetic loopback HTTP only. Save the test below as `internal/telemetry/profiles/review_truncation_test.go`; it reuses this PR's existing `testCfg` and `dataframeBody` fixtures. Run `GOTOOLCHAIN=go1.27.1 go test -mod=readonly -p 1 -race ./internal/telemetry/profiles -run TestReviewTruncatedRankings -count=1 -timeout=30s -v`.

Expected: capped responses disclose truncation. Actual: `top` and `summary` fail with `total_samples=10000 truncated=<nil>`; the `flamegraph` positive control passes. The comparison base `c3f7734cdb6d5c894f4f9f99dbbfa641d1bda9f5` has no profiling package, so there is no equivalent old handler to execute. The fixture follows the dashboard's documented sample-descending query and row limit; no live backend claim.

```go
package profiles

import (
 "context"
 "encoding/json"
 "fmt"
 "net/http"
 "net/http/httptest"
 "testing"
 "github.com/modelcontextprotocol/go-sdk/mcp"
)

// Returned stack rows are a bounded subset, so both derived tools must disclose it.
func TestReviewTruncatedRankings(t *testing.T) {
 for _,kind:=range []string{"flamegraph","top","summary"} {t.Run(kind,func(t *testing.T){
  // 1000 heavier distinct stacks precede 1001 lighter stacks of the same hot leaf.
  // Server ordering is sample-descending; the omitted leaf contributes 9009/19009 samples.
  rows:=make([]map[string]any,0,2001)
  for i:=0;i<1000;i++ {rows=append(rows,map[string]any{"StackHash":fmt.Sprint(i),"Frames":fmt.Sprintf("leaf%d\x1fmain",i),"samples":10})}
  for i:=0;i<1001;i++ {rows=append(rows,map[string]any{"StackHash":fmt.Sprint(1000+i),"Frames":fmt.Sprintf("hottest\x1fcaller%d",i),"samples":9})}
  srv:=httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter,r *http.Request){
   var req struct {Pipeline []map[string]any `json:"pipeline"`};json.NewDecoder(r.Body).Decode(&req)
   limit:=int(req.Pipeline[len(req.Pipeline)-1]["limit"].(float64));if limit>len(rows){limit=len(rows)}
   w.Write(dataframeBody(rows[:limit]...))
  }));defer srv.Close()
  cfg:=testCfg(srv.URL);var res *mcp.CallToolResult;var err error
  switch kind {case "flamegraph":res,_,err=NewGetFlamegraphHandler(srv.Client(),cfg)(context.Background(),nil,GetFlamegraphArgs{Service:"api"})
  case "top":res,_,err=NewGetTopFunctionsHandler(srv.Client(),cfg)(context.Background(),nil,GetTopFunctionsArgs{Service:"api"})
  case "summary":res,_,err=NewGetProfileSummaryHandler(srv.Client(),cfg)(context.Background(),nil,GetProfileSummaryArgs{Service:"api"})}
  if err!=nil||res.IsError{t.Fatalf("%+v %v",res,err)}
  var got map[string]any;json.Unmarshal([]byte(res.Content[0].(*mcp.TextContent).Text),&got)
  if got["total_samples"]!=float64(10000){t.Fatalf("fixture no longer capped: %v",got["total_samples"])}
  if got["truncated"]!=true {t.Fatalf("bounded 1000/2001 stacks reported without truncation: total_samples=%v truncated=%v; full fixture total=19009",got["total_samples"],got["truncated"])}
 })}
}


```

nishantmodak (changes requested):
Reviewed the complete 31-file change at `f59f90b8ad2c0b352ba3ebf79ea4d474c0a58d02`, including all four handlers, query/response adapters, tree and recursive-frame folding, resource/tenant scoping, time ranges, descriptions, registration/toolsets and existing tests. Compared the query contract with the dashboard implementation and read existing discussions/CI.

One required correction: the top-functions and summary tools silently treat a capped stack subset as the whole profile, potentially hiding the actual hottest function. The inline thread includes a copyable regression, reproduced twice; the flamegraph control passes.

Local Go 1.27.1/race validation: existing profiling and toolset tests, dump-tools tests, actual profiles/logs registration checks, 100 seeded conservation cases, nine HTTP-status/scope cases, malformed JSON/empty response, optional timestamp failure and cancellation checks. Only the two new truncation assertions fail. No live backend/customer data or full local integration suite; two opt-in live tests skipped. Current-head Run Tests/CodeQL checks pass separately. No other actionable blockers found.

internal/telemetry/profiles/get_top_functions.go
prathamesh-sonpatki: Done — `get_top_functions` and `get_profile_summary` now set `truncated` when the flamegraph row cap is hit (same signal as `get_flamegraph`), qualify summary text as covering the returned stack subset, and drop the full-profile `total_samples` guarantee from the tool descriptions. Added `TestTruncatedRankingsDiscloseCap` covering flamegraph/top/summary.

nishantmodak (approved):
Reviewed the complete PR at `72cee5abe270fa762fdd64b3f779983d96af1930`, including the four profiling tools, query/response adapters, frame folding, resource/tenant scoping, toolsets/registration, descriptions and merged shared-code interactions. No remaining actionable blockers found.

The earlier truncation finding is fixed. I reran the exact attached `TestReviewTruncatedRankings` assertion twice on this head: flamegraph, top-functions and summary all pass. The summary and descriptions now qualify capped totals/rankings.

Local Go 1.27.1 validation with `-race -p 1`: profiling/toolset tests, dump-tools tests, 100 seeded conservation cases, HTTP error/filter and cancellation probes, 0/1/999/1000-row boundaries, deterministic concurrent tenant scoping, and actual profiles/logs registration checks all pass. Current-head GitHub Run Tests and CodeQL checks also pass. Synthetic HTTP only; two opt-in live tests skipped, and the full integration suite was not repeated locally.

PR #215, opened 2026-08-14
ENG-1735: steer create_dashboard as net-new, refine with update_dashboard

prathamesh-sonpatki:
## Summary

Agents call `create_dashboard` repeatedly while refining panels, producing a duplicate dashboard per edit. Create once for a net-new dashboard; panel/layout/query changes then go to `update_dashboard` with the returned id.

Soft steer only — no server-side block on a second create. Two tools are kept rather than merged into an upsert, so an approval prompt can still distinguish "create new" from "mutate existing".

## Changes

- `create_dashboard.md` / `update_dashboard.md`: net-new vs refine copy
- `create.go`: on successful create, append a second `TextContent` naming `update_dashboard`, the id, and the full-replacement body contract. `Content[0]` stays raw API JSON, so existing consumers are unaffected.
- Tests: description contract (`dump-tools` served text == embed), nudge with and without an id

## Test plan

- [ ] `go test ./internal/prompts/ ./internal/dashboards/ . -count=1`
- [ ] "create an error logs dashboard" → one `create_dashboard`, refinements via `update_dashboard`
- [ ] Create result still parses as JSON in `Content[0]` and includes `dashboard.id`


prathamesh-sonpatki:
## Codex review (`codex review --base main`)

The implementation and tests pass. One finding:

**[P2] Scope the create ban to the dashboard already created** — `internal/prompts/descriptions/create_dashboard.md:3`

When a user requests two distinct net-new dashboards in one turn, the sentence keyed to any ID returned this turn forbids `create_dashboard` after the first call, even though the second dashboard also requires creation. The result nudge repeats this ambiguity. Scope the prohibition to re-creating or refining the *same* dashboard (panel/layout/query tweaks), not to a second net-new dashboard.

No other defects flagged. HITL unchanged and two-tool split look intentional.

prathamesh-sonpatki:
## Thermo-nuclear review

### Correctness / security
**No medium-or-higher findings.** HITL unchanged, two tools stay separate, `Content[0]` stays raw API JSON, empty id does not invent a nudge, no hard block on a second create. Codex’s “global create ban” is prompt steering matching ENG-1735, not a server gate.

### Code quality
**Do not approve as-is on structure.** `internal/writeintent` is a six-file framework around one `fmt.Sprintf` and phrase lists for a single write pair. `ForbiddenCreatePhrases` still hardcodes `list_dashboards`. `Annotate` is a thin mutating wrapper; `dashboardIDFromResponse` runs twice. The “future `create_alert` table row” is not in product.

**Code judo:** inline the nudge in `internal/dashboards/create.go` (or `result.go`, which already extracts the id). Keep phrase checks as `[]struct{phrase, reason}` in `prompts_test.go` like `TestGetLogsDescriptionCriticalRules`. Delete `writeintent`.

faiz9068 (approved):

PR #216, opened 2026-08-15
fix(mcp): query get_change_events filters against stored metric labels

prathamesh-sonpatki:
## Summary

`get_change_events` filtered on `service_name`, `env`, and `event_type` — none of which are the labels change events are stored with (`service`, `deployment_environment`, `event_name`). A filtered call returned nothing while the same call without filters returned the events.

The query is a single selector on the stored labels. Tool parameter names are unchanged. Filter values are escaped before interpolation, and `available_event_names` applies the query's exclusion matchers so every name it offers is one a filter can match.

## Test plan

- [x] `go test ./internal/change_events/`
- [x] `go test . -run TestDumpTools`
- [x] Verified against live data: filtered counts agree with unfiltered, and a value containing a quote produces a well-formed query instead of an upstream 400.

### DIFF
--- a/internal/change_events/change_events.go
+++ b/internal/change_events/change_events.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -69,34 +70,7 @@ func NewGetChangeEventsHandler(client *http.Client, cfg models.Config) func(cont
 			return nil, nil, fmt.Errorf("failed to fetch available event names: %w", err)
 		}
 
-		// Build label filters for the Prometheus query
-		var labelFilters []string
-
-		if args.ServiceName != "" {
-			labelFilters = append(labelFilters, fmt.Sprintf(`service_name="%s"`, args.ServiceName))
-		}
-
-		if args.Env != "" {
-			labelFilters = append(labelFilters, fmt.Sprintf(`env="%s"`, args.Env))
-		}
-
-		// Use event_name parameter directly - the AI should provide the exact event type
-		if args.EventName != "" {
-			labelFilters = append(labelFilters, fmt.Sprintf(`event_type="%s"`, args.EventName))
-		}
-
-		// Add default filters to exclude backup and rehydration events
-		labelFilters = append(labelFilters, `event_name!~"cold_storage_logs_backup|cold_storage_logs_backup_endtime|cold_storage_logs_backup_time_taken_in_sec|manual_rehydration_event"`)
-		labelFilters = append(labelFilters, `l9_event_name!~"last9_scheduled_search"`)
-
-		// Build the filter string
-		var filterStr string
-		if len(labelFilters) > 0 {
-			filterStr = "{" + strings.Join(labelFilters, ",") + "}"
-		}
-
-		// Build PromQL query for change events
-		promql := fmt.Sprintf("last9_change_events%s", filterStr)
+		promql := buildChangeEventsPromQL(args.ServiceName, args.Env, args.EventName)
 
 		// Make range query to get change events over time
 		resp, err := utils.MakePromRangeAPIQuery(ctx, client, promql, startTimeParam, endTimeParam, cfg)
@@ -147,10 +121,71 @@ func NewGetChangeEventsHandler(client *http.Client, cfg models.Config) func(cont
 	}
 }
 
-// fetchAvailableEventNames fetches all available event_name values from the last9_change_events metric
+const (
+	changeEventsMetric     = "last9_change_events"
+	backupEventExclude     = `event_name!~"cold_storage_logs_backup|cold_storage_logs_backup_endtime|cold_storage_logs_backup_time_taken_in_sec|manual_rehydration_event"`
+	scheduledSearchExclude = `l9_event_name!~"last9_scheduled_search"`
+)
+
+// buildChangeEventsPromQL maps MCP-canonical params (service_name, env,
+// event_name) onto the labels last9_change_events actually stores.
+//
+// Stored series use service / deployment_environment / event_name. Some
+// tenants also copy the MCP names (service_name / env / event_type). Each
+// constraint matches the primary label, or the alias only when the primary
+// is absent — an unconditional OR would false-positive when both labels
+// exist with different values.
+func buildChangeEventsPromQL(serviceName, env, eventName string) string {
+	sels := [][]string{{backupEventExclude, scheduledSearchExclude}}
+	sels = expandPrimaryOrAlias(sels, "service", "service_name", serviceName)
+	sels = expandPrimaryOrAlias(sels, "deployment_environment", "env", env)
+	sels = expandPrimaryOrAlias(sels, "event_name", "event_type", eventName)
+
+	out := make([]string, 0, len(sels))
+	for _, f := range sels {
+		out = append(out, changeEventsMetric+"{"+strings.Join(f, ",")+"}")
+	}
+	return strings.Join(out, " or ")
+}
+
+func expandPrimaryOrAlias(sels [][]string, primary, alias, value string) [][]string {
+	if value == "" {
+		return sels
+	}
+	out := make([][]string, 0, len(sels)*2)
+	for _, s := range sels {
+		out = append(out, append(append([]string{}, s...), fmt.Sprintf(`%s="%s"`, primary, value)))
+		out = append(out, append(append([]string{}, s...), fmt.Sprintf(`%s=""`, primary), fmt.Sprintf(`%s="%s"`, alias, value)))
+	}
+	return out
+}
+
+// fetchAvailableEventNames unions event_name and event_type values so
+// discover-then-filter works for both stored shapes.
 func fetchAvailableEventNames(ctx context.Context, client *http.Client, startTime, endTime int64, cfg models.Config) ([]string, error) {
-	// Use the label values API to get all event_name values
-	resp, err := utils.MakePromLabelValuesAPIQuery(ctx, client, "event_type", "last9_change_events", startTime, endTime, cfg)
+	unique := make(map[string]struct{})
+	for _, label := range []string{"event_name", "event_type"} {
+		names, err := fetchEventNamesForLabel(ctx, client, label, startTime, endTime, cfg)
+		if err != nil {
+			return nil, err
+		}
+		for _, name := range names {
+			if name == "" {
+				continue
+			}
+			unique[name] = struct{}{}
+		}
+	}
+	eventNames := make([]string, 0, len(unique))
+	for name := range unique {
+		eventNames = append(eventNames, name)
+	}
+	sort.Strings(eventNames)
+	return eventNames, nil
+}
+
+func fetchEventNamesForLabel(ctx context.Context, client *http.Client, label string, startTime, endTime int64, cfg models.Config) ([]string, error) {
+	resp, err := utils.MakePromLabelValuesAPIQuery(ctx, client, label, "last9_change_events", startTime, endTime, cfg)
 	if err != nil {
 		return nil, fmt.Errorf("failed to query event names: %w", err)
 	}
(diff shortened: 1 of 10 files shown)


internal/change_events/change_events.go
```diff
@@ -147,10 +121,71 @@ func NewGetChangeEventsHandler(client *http.Client, cfg models.Config) func(cont
...
+	return out
+}
+
+// fetchAvailableEventNames unions event_name and event_type values so
+// discover-then-filter works for both stored shapes.
 func fetchAvailableEventNames(ctx context.Context, client *http.Client, startTime, endTime int64, cfg models.Config) ([]string, error) {
-	// Use the label values API to get all event_name values
-	resp, err := utils.MakePromLabelValuesAPIQuery(ctx, client, "event_type", "last9_change_events", startTime, endTime, cfg)
+	unique := make(map[string]struct{})
+	for _, label := range []string{"event_name", "event_type"} {
```
faiz9068: `available_event_names` unions `event_type`, but nothing stamps an `event_type` label — every change event carries `event_name`, set unconditionally from a required field at ingest.

So an `event_type` value surfaced here can never be matched by the filter below: the alias branch requires `event_name=""`, which never holds. The description tells the model to filter using an exact name from this list, so such a value returns a confident zero — the same failure this PR fixes, in a narrower spot.

Live, label-values for `event_type` returns `[]`, so the union contributes nothing today. Suggest reading `event_name` only, which also drops a round trip per call.

internal/change_events/change_events.go
```diff
@@ -147,10 +121,71 @@ func NewGetChangeEventsHandler(client *http.Client, cfg models.Config) func(cont
...
+}
+
+func expandPrimaryOrAlias(sels [][]string, primary, alias, value string) [][]string {
+	if value == "" {
+		return sels
+	}
+	out := make([][]string, 0, len(sels)*2)
+	for _, s := range sels {
+		out = append(out, append(append([]string{}, s...), fmt.Sprintf(`%s="%s"`, primary, value)))
+		out = append(out, append(append([]string{}, s...), fmt.Sprintf(`%s=""`, primary), fmt.Sprintf(`%s="%s"`, alias, value)))
```
faiz9068: Values are interpolated unescaped — a value containing `"` or `\` produces a malformed selector and an opaque upstream 400.

The repo already has `escapePromQLLabel` (`internal/apm/databases.go:545`), used at the parallel sites in `deviations_query.go:209`/`212`. Pre-existing behavior, but this PR rewrote these lines and multiplied the interpolation sites. Wrapping `value` once at the top of the function covers every branch.

faiz9068 (commented):
Reviewed the mapping fix and verified it against live data (own org, 7d), this branch vs `main`:

- unfiltered: 37 / 37
- filters that returned 0 on main now return events (env, event_name, and a service that stores only `service`)
- a service carrying both spellings: 20 / 20, 20 unique series signatures — no double-count
- bogus filter values: 0 on both

Mechanically the fix is right: `primary=""` matches absent-or-empty, so the two branches per filter are disjoint; `or` preserves `__name__` and the full label set, so the response `metric` map is unaffected; the exclusion constants carry into every branch. The expansion caps at 8 branches (three params), which is fine — the `and`-factored alternative is 7 selectors at n=3 with a subtler correctness argument, so no win.

Comments inline. Two I'd fix before merge: the `event_type` union in `available_event_names` (surfaces names no filter can match), and unescaped label interpolation. One needs a decision rather than an edit — the both-spellings-disagree narrowing.

Also: `CHANGELOG.md:12` documents the `event_type` union, so it changes with that fix. And `get_change_events.md:16` points at `did_you_mean`, which resolves over `trace_*` metrics where labels really are `service_name`/`env` — a resolved value may not exist as `service`/`deployment_environment` here (pre-existing, unchanged line, so no inline).

internal/change_events/change_events.go
```diff
@@ -147,10 +121,71 @@ func NewGetChangeEventsHandler(client *http.Client, cfg models.Config) func(cont
...
+	scheduledSearchExclude = `l9_event_name!~"last9_scheduled_search"`
+)
+
+// buildChangeEventsPromQL maps MCP-canonical params (service_name, env,
+// event_name) onto the labels last9_change_events actually stores.
+//
+// Stored series use service / deployment_environment / event_name. Some
+// tenants also copy the MCP names (service_name / env / event_type). Each
+// constraint matches the primary label, or the alias only when the primary
+// is absent — an unconditional OR would false-positive when both labels
```
faiz9068: Two notes on this rationale:

1. `or` de-duplicates identical label sets, so an unconditional OR wouldn't double-count. What the `primary=""` guard actually buys is precedence, not de-duplication.
2. It's also a narrowing: a series carrying `service="A"` and `service_name="B"` matches neither branch when filtering `service_name="B"` — the old code matched it. The label whose name equals the param is ignored whenever the primary is present.

Defensible as a deliberate precedence rule; worth saying so in those terms and confirming the narrowing is intended.

internal/change_events/change_events.go
```diff
@@ -147,10 +121,71 @@ func NewGetChangeEventsHandler(client *http.Client, cfg models.Config) func(cont
...
+// exist with different values.
+func buildChangeEventsPromQL(serviceName, env, eventName string) string {
+	sels := [][]string{{backupEventExclude, scheduledSearchExclude}}
```
faiz9068: Naming: these aren't selectors — a selector is `last9_change_events{…}`, built at line 144. Each element is the matcher set for one future branch, so the name sits one level off and `[][]string` doesn't recover it.

Same area: `out` means two different things (`[]string` of selectors at :144, `[][]string` of matcher sets at :155), and `s`/`f` name the same concept in the two loops.

A `type matcherSet []string` plus `branches`/`selectors` would carry it, and would let the doc comment above drop its "what" half and keep only the non-obvious "why".

internal/change_events/change_events_test.go
```diff
@@ -99,3 +99,47 @@ func TestGetChangeEventsArgs_UsesCanonicalNames(t *testing.T) {
 		}
 	}
 }
+
+func TestBuildChangeEventsPromQL(t *testing.T) {
```
faiz9068, line 103: These pin branch order and matcher order within a branch, neither of which is semantic — reordering the three `expandPrimaryOrAlias` calls is a pure refactor that breaks all five cases.

Two gaps: no three-filter case, so the 8-branch path (the one with the most room to go wrong) is untested; and no value containing a quote. Keeping one golden string and asserting the rest on the set from `strings.Split(q, " or ")` would survive refactors and still catch regressions.

internal/prompts/descriptions/get_change_events.md
```diff
@@ -2,7 +2,7 @@ Change events from the `last9_change_events` Prometheus metric (deployments, con
 
 Response: `available_event_names` (valid filter values), `change_events` (timeseries with metric labels + timestamp/value pairs), `count`, `time_range`.
 
-Each event has `metric` (labels: service_name, env, event_type, message, …) and `values` (timestamp/value pairs).
+Each event has `metric` (labels: service, deployment_environment, event_name, message, …) and `values` (timestamp/value pairs). Pass tool params `service_name` and `env` — they map onto those stored labels (and `service_name`/`env` when present).
```
faiz9068: The parenthetical repeats the param names it just named, so it's ambiguous which spelling is the param and which is the stored label. It also omits `event_name`, which has the same mapping.

Suggest: "Filter with the tool's own param names (`service_name`, `env`, `event_name`) — the server maps them onto whichever spelling the series stores."

internal/change_events/change_events.go
faiz9068: Removed `event_type` entirely — from discovery and from the query. `event_name` is required at ingest, so an alias arm gated on `event_name=""` could never match.

Discovery now reads `event_name` alone and applies the query's exclusion matchers. Without them it offered names that every filtered call then rejects.

internal/change_events/change_events.go
faiz9068: Fixed. Extracted the existing helper to `utils.EscapePromQLLabel` and applied it to every interpolated value. The `apm` copy calls it too, so there is one implementation rather than a third.

internal/change_events/change_events.go
faiz9068: Removed `event_type` entirely — from discovery and from the query. `event_name` is required at ingest, so an alias arm gated on `event_name=""` could never match.

Discovery now reads `event_name` alone and applies the query's exclusion matchers; without them it offered names that every filtered call then rejects.

internal/change_events/change_events.go
faiz9068: Fixed. Extracted the existing helper to `utils.EscapePromQLLabel` and applied it to every interpolated value. The `apm` copy calls it too, so there is one implementation rather than a third.

internal/change_events/change_events.go
faiz9068: Both correct. Resolved by dropping the two-arm expansion entirely — the query is now a single selector, so there is no precedence rule and no narrowing.

Label data showed no series carrying an MCP spelling without the stored one, so the alias arm never fired.

internal/change_events/change_events.go
faiz9068: Moot now — the expansion is gone. One flat `[]string` of matchers, one selector.

internal/change_events/change_events_test.go
faiz9068: Rewritten as a table test asserting exact queries, with a three-filter case and values containing a quote and a backslash. Nothing pins a non-semantic order any more.

internal/prompts/descriptions/get_change_events.md
faiz9068: Applied.

faiz9068 (commented):
Re-reviewed the two new commits (`9158c1d`, `7cf5ede`). All six earlier findings are addressed, and two went further than asked. CI green on head.

**Verified resolved**

1. `event_type` union — gone; discovery reads `event_name` only, and now carries the query's exclusions so the list can't offer an excluded name. Good addition.
2. Escaping — fixed and generalized to `utils.EscapePromQLLabel`, with the four `internal/apm` call sites migrated. Verified live: a `service_name` of `x"} or up{` returns `400 Invalid query` on `main` and a clean `count: 0` on this head.
3. Alias precedence — resolved by removing the alias entirely (see the note below).
4. Naming — `matchers`/`addFilter`, no more `[][]string`, and the comment is down to the non-obvious part. Reads cleanly now.
5. Description wording — adopted; the `event_name` jsonschema saying "event type" is fixed too.
6. Tests — table-driven, with the all-three case and the escape/injection cases that were missing.

**On dropping the alias entirely.** Stronger than what I'd suggested, and I couldn't find data that it breaks. Across two tenants' `last9_change_events`: zero series carry `service_name` without `service`, or `env` without `deployment_environment` (checked with `last_over_time(...[7d])`; the 674-series baseline confirms the probe isn't vacuous). One caveat worth a decision, not a blocker: the change-events API docs example advertises `attributes: {env: staging}`. A producer that followed that example literally stores `env` and is no longer matchable by the `env` filter. Either update that example to the stored spelling, or accept it knowingly.

**Merge-readiness:** CI green on `7cf5ede`, mergeable, no conflicts. Nothing blocking from my side.

internal/change_events/change_events.go
```diff
@@ -147,10 +121,51 @@ func NewGetChangeEventsHandler(client *http.Client, cfg models.Config) func(cont
...
+func addFilter(matchers []string, label, value string) []string {
+	if value == "" {
+		return matchers
+	}
+	return append(matchers, fmt.Sprintf(`%s="%s"`, label, utils.EscapePromQLLabel(value)))
+}
+
+// Same exclusions as the query: a name this list offers must be one a filter
+// can actually match. event_name only — no other label carries an event
+// identity a filter can match.
```
faiz9068: This invariant doesn't hold at window boundaries, and it's newly reachable now that discovery returns names at all (on `main` it returned `[]`, so it never offered a bad one).

The label-values API resolves names over a coarser window than the range query, so it offers names whose samples fall outside the requested range. Measured on this head, same call, varying only the lookback:

```
1d    unfiltered=0    names=0   names returning zero: []
7d    unfiltered=13   names=8   names returning zero: [feature_flag_updated incident_note]
30d   unfiltered=39   names=8   names returning zero: []
```

At 7d the discovery list has 8 entries but the range response contains only 6 distinct `event_name` values — the two extra are offered, and the description tells the model to "filter with the exact name from that list," so it picks one and gets the same confident zero this PR is fixing.

Cheapest exact fix: for the unfiltered call the range response already carries every matching series, so `available_event_names` can be derived from it — same window by construction, and it drops the extra upstream call. Otherwise the list needs a caveat that a name may fall outside the window.

faiz9068 (approved):

PR #218, opened 2026-08-17
fix(logs): surface plain-text Body shape and un-suppress zero-count sanity

prathamesh-sonpatki:
Two fixes for the same failure class: a log query that returns a wrong number without anything signalling it.

## The problem

`get_log_attributes_for_pipeline` classifies a Body as JSON, then logfmt, then plaintext-with-severity-token. A Body matching none of those — plain text like `<ts> PUT [200] /path 4.4ms` — fell through to a bare `return nil`. The only Body entry the model saw was the catalog's `{"$eq":["Body","<value>"]}`, which says nothing about shape.

Models filled the gap by guessing, in two ways:

1. `{"type":"parse","parser":"json","field":"Body"}` — matches nothing, returns `count: 0` with `"status":"success"`. A silent zero.
2. `(?P<status_code>[0-9]{3})` — unanchored, so it captures the leading **timestamp** digits, not the status. Returns a confidently wrong non-zero count.

Both are worse than an error, because both look like answers.

## Fix A — report the Body's shape

A Body with no recognized structure now yields one entry with `source:"body"`, a redacted `sample_body` sample line, and a one-stage `$regex`-on-Body hint (no parse stage) to anchor against.

The sample costs no extra query — it is the same sample `sample_coverage` is already computed from, previously discarded at `return nil`.

One subtlety worth review attention: the plaintext entry has to survive the merge in `discoverLogAttributes`, which drops body-derived entries whose `filter_field` duplicates an indexed one. The series catalog always exposes `body` → `Body` for exactly these services, so the fallback was a *guaranteed* duplicate and was silently dropped — the first version of this fix was completely inert in production while its unit tests passed, because the test fixtures had no `body` field in the series names. It now enriches the existing indexed entry in place. `TestGetLogAttributesForPipeline_PlaintextFallbackSurvivesIndexedBodyDup` covers this and was confirmed to fail against the unfixed merge loop.

## Fix B — stop suppressing the guardrail on zero

`AppendCountSanity` opened with `if matchedCount <= 0 { return response }`, so a zero count — the outcome most likely to be wrong — was the single case that got no `l9_sanity` block. The guardrail switched itself off precisely when it was needed.

A zero from a pipeline containing a `parse` stage now carries a note pointing at `sample_body` to confirm the real shape before retrying. Zero with no parse stage stays untouched (a plain filter matching nothing is unremarkable), and every non-zero path is byte-for-byte unchanged. No PromQL baseline fetch on the zero path — there is nothing to compare a zero against.

## Verification

`go build ./...` and `go test ./...` green.

Behavioural check via an instrumented agentic probe against the live server — same prompt, 4 runs per build, inspecting the queries the model actually wrote. Ground truth measured independently at **397** for the window:

| | main | this branch |
|---|---|---|
| `parser:"json"` guesses | 5 | **0** |
| `get_logs` calls | 10 | **4** (one per run) |
| answers | 392, 394, 395, **163** | 397, 388, 393, 399 |
| wrong answers | 1 of 4 | **0 of 4** |

Every run on this branch goes discovery → one anchored query → done. The 388–399 spread is window drift across runs minutes apart (a control query moved 394 → 396 → 397 over the same period), not error.

## What this does NOT show

The `scoped_log_attribute_discovery` agentic suite scores **2/6 on main and 3/6 here, and that one-point difference is flake, not signal** — slad-003 scored 2/2, then 0/2, then 3/4 across identical code. Both builds were run against the same updated case file, so the comparison is clean; the suite simply cannot see this fix.

The reason is structural: slad-001 is graded by `expect_nonzero`, and the wrong answer of 163 **is** non-zero. The rubric passes a 59%-undercount as long as it isn't zero. That is a gap in the eval, not evidence the fix does nothing — hence the probe above, which measures query shape and answer accuracy directly. Worth a follow-up in `last9-mcp-evals` to give that case a count tolerance or an anchored-pattern rule.

## Notes

- Independent of #217 (the v0.15.1 release PR, which only touches `package.json` + changelog headings). Both changes here are additive — an entry where discovery previously returned nothing, a block where none was emitted — so no existing response shape changes.
- `sample_body` is redacted through the existing `SanitizeUpstreamBody` (URLs, credentials, 512-byte cap). Note it does not strip bare request paths, so `sample_body` inherits whatever identifiers sit in the body — the same data `get_logs` and `get_service_logs` already return verbatim. Considered and accepted as-is: the exposure is not new, and path-redacting would break the feature's purpose (the path shape is often exactly what a regex needs to anchor against). If revisited, the narrow version is redacting only tenant-identifying path segments, not paths wholesale.
- All test fixtures use synthetic bodies.

---

## Review round (7 reviewers, run `20260817-092403`)

Two additional commits address findings from a seven-reviewer pass. Two were blocking:

**Enrichment hijacked an indexed `level` entry** (correctness + testing + adversarial, independently). The merge exception keyed on `SampleBody != ""`, which stopped identifying the plaintext fallback once the severity-token entry also set that field. A service with an indexed field named `level` whose Body also carries a severity token had its directly-filterable entry overwritten with a Body parse hint — the dup check fires before severity-family suppression. Now scoped to `FilterField == "Body"`; the regression test was confirmed to fail against the old condition.

**The zero-count guardrail excluded the shape this PR recommends** (adversarial). The hint emitted for a plain-text Body is parse-free by construction, while both zero paths gated on `hasParseStage` — so a model following our own hint into a zero got no guardrail, and a verbatim `<your-pattern>` copy is a valid regex that matches nothing and returns a confident 0. The gate now covers pipelines that parse *or* filter Body. Both zero checks also moved ahead of a single-service gate they never needed.

Also fixed: `sample_body` was described as "REDACTED" when `SanitizeUpstreamBody` strips only URLs and obvious credentials (bare paths, emails, IPs, `Set-Cookie`, prefix-format tokens all survive) — the false label is the risk, since a model told a string is sanitized echoes it more freely; it is now described accurately and marked untrusted. On the severity branch `sample_body` was the first non-blank line rather than one matching the emitted pattern, so a sample could contradict its own hint. `firstNonEmptySampleBody` returned `""` when the first non-blank line sanitized away.

### Verification

Live wire: the zero note fires for a parse-free Body regex and for the grouped empty-result shape, and stays absent when the pipeline neither parses nor touches Body. `log_query` evals 50/50 on this branch, matching master. Agentic suite 2/6, identical to master's 2/6 and the same per-case profile (the difference in earlier runs was slad-003, which has scored 2/2, 0/2, 3/4, 1/2 and 0/2 across identical code).

### Known open items — deliberately not fixed here

- **Sanitizer fidelity.** `SanitizeUpstreamBody` was written for upstream error bodies; reused for a value whose contract is byte fidelity. Multi-line bodies collapse to one line, the credential regex removes the key along with the value, and 512-byte truncation is undocumented — while the description says "anchor your pattern to that exact shape". Needs a decision: stop reusing the helper here, disclose the full transform set, or emit per-sample transform notes.
- **Representativeness.** One sample line, no coverage figure. A regex anchored to an outlier in a mixed-format service undercounts silently, and a low-but-nonzero ratio produces an *empty* `l9_sanity` note — reading as validated.
- **Genuine zeros.** The note fires for any zero from any pipeline touching Body, and dead-ends when the service emitted no logs at all (no `sample_body` exists to inspect). The `physical_index_service_count` baseline already computed for the nonzero path would settle it.
- `AppendCountSanity` is reachable only because `shouldChunkGetLogsQuery` returns false for aggregates; extending chunking to aggregates would silently delete this guardrail, untested.


internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -56,8 +70,133 @@ const (
...
+	if sampleBodyURLPattern.MatchString(s) {
+		s = sampleBodyURLPattern.ReplaceAllString(s, "[redacted-url]")
+		addNote("url-redacted")
+	}
+
+	// Redact credential VALUES while preserving the key + separator, so the
+	// model can still see that a token/secret param existed at all.
+	before := s
+	s = sampleBodyCredentialPattern.ReplaceAllString(s, "$1$2[redacted]")
+	s = sampleBodyBearerPattern.ReplaceAllString(s, "$1 [redacted]")
```
faiz9068: Credential pass before bearer pass leaks the token.

`sampleBodyCredentialPattern`'s value class is `[^"',}\s]+`, so it stops at the first space. On `Authorization: Bearer <jwt>` it consumes only the word `Bearer`; `sampleBodyBearerPattern` then has nothing left to anchor on and never fires:

```
IN : auth failed Authorization: Bearer eyJhbGciOiJIUzI1NiJ9.PAYLOAD.sig for user
HERE: auth failed Authorization: [redacted] eyJhbGciOiJIUzI1NiJ9.PAYLOAD.sig for user
SanitizeUpstreamBody: auth failed [redacted-credential] for user
```

`authorization=Bearer abc123` goes the same way. `SanitizeUpstreamBody` (upstream_http.go:24-26) applies bearer first and gets it right — the fork inverted it. `addNote("credential-redacted")` still fires, so `sample_notes` claims a redaction that didn't happen, which is the part that worries me most: this is the one path that deliberately echoes raw log lines to a model.

Swapping the two lines fixes it (`Authorization: [redacted] [redacted]` is fine as output). Tests currently only cover `token=abc` — worth adding `Bearer` and `Basic` cases.

Separately: the comment above justifies forking for whitespace/key-preservation, which is fair, but the pattern set and ordering needn't be forked. Exporting the three vars from `internal/utils` would stop the two sanitizers drifting again.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `l9_sanity` is no longer suppressed on a zero count. `AppendCountSanity` returned early whenever the matched count was zero, so the most suspicious outcome was the one case that got no guardrail. A zero count from a pipeline containing a `parse` stage now carries a note pointing at `sample_body` to confirm the real Body shape before retrying. A zero count with no `parse` stage is still left untouched, and non-zero behaviour is unchanged.
```
faiz9068: Wrong section, and the text predates the review round.

Both bullets are under `## [0.15.1] - 2026-08-16`, which is tagged and released — this branch's base commit is `chore: release v0.15.1 (#217)`. As written the changelog claims the fix shipped in 0.15.1. They belong under `## [Unreleased]`, which is currently empty.

Content drifted too:
- "a redacted `sample_body` sample line" — the field is `sample_bodies` (up to 3 lines), and a review round specifically established it is *not* PII-redacted. `sample_body` no longer exists in the code; repeating "redacted" in release notes undoes that correction for anyone who only reads the changelog.
- "A zero count from a pipeline containing a `parse` stage ... with no `parse` stage is still left untouched" — the gate is parse **or** filters-Body now. The parse-free `$regex` shape this PR recommends is the interesting case.
- No mention of `service_log_volume`, which is a new user-visible key in the `l9_sanity` block.

faiz9068 (commented):
Two blocking items from a read-through;

1. `sampleBodyForModel` runs its two redaction passes in the opposite order to `SanitizeUpstreamBody`, and a `Bearer` token survives as a result.
2. Both changelog bullets landed under the already-tagged `0.15.1` section, and their wording predates the review round.

Details inline.

internal/telemetry/logs/attributes_for_pipeline.go
faiz9068: Confirmed and fixed. Reproduced first — the failing test printed your trace exactly:

```
IN : auth failed Authorization: Bearer eyJhbGciOiJIUzI1NiJ9.PAYLOAD.sig for user
OLD: auth failed Authorization: [redacted] eyJhbGciOiJIUzI1NiJ9.PAYLOAD.sig for user
NEW: auth failed Authorization: [redacted] for user
```

Bearer now runs first, matching `SanitizeUpstreamBody`. I also let the credential pattern absorb an optional `(?:bearer|basic)\s+` scheme, so a keyed credential collapses to a single `[redacted]` rather than the `[redacted] [redacted]` a bare swap produces — and that covers `Basic`, which had no pattern at all and leaked its base64 the same way.

One deliberate gap: I did not add `basic` to the bare scheme pattern alongside `bearer`. `basic` is ordinary prose, so `basic auth failed for user bob` would become `basic [redacted] for user bob` — corrupting the byte fidelity these samples exist to provide. `Bearer` is safe bare because it is effectively never a prose word. So bare `Basic <base64>` with no preceding key is still unredacted. Say the word and I'll add it with a base64-shaped value class instead of `\S+`.

Tests: one table-driven case covering `Authorization: Bearer`, `authorization=Bearer`, `Authorization: Basic` and bare `Bearer`, plus a prose-`basic` case asserting the line comes back byte-identical.

On the export suggestion — taken in part, and I want to flag where I went a different way. Exporting the three compiled vars doesn't work: `sampleBodyCredentialPattern` needs capture groups to preserve the key and separator, which is the reason for the fork in the first place, and reshaping the `utils` one to match would change `SanitizeUpstreamBody`'s output on every 400 body. Instead both patterns are now built from an exported `utils.CredentialKeyAlternation`, since the key list is the drift with actual consequences — someone adding a key to one list and not the other. Ordering can no longer drift silently because it is now identical in both and commented. If you want the full three-var export regardless, that's a change to `SanitizeUpstreamBody` behaviour and I'd rather have your sign-off than assume.

Commits are local on the branch, not pushed yet.

CHANGELOG.md
faiz9068: Confirmed on all three points. Both bullets moved out of the tagged `0.15.1` section into `[Unreleased]` under a new `### Fixed`.

Rewording, each checked against the code rather than the PR description:

- `sample_body` does not exist — it is `sample_bodies` (up to three lines), with `sample_notes` alongside. Now described as redacting URLs and known credential values only, explicitly **not** PII-redacted, and to be treated as untrusted log content.
- Gate corrected to parse **or** filters-`Body`, with the parse-free `$regex` shape called out as the case that matters, since that is the shape the tool itself recommends.
- `service_log_volume` documented as a new key, framed by what it is for — separating a genuine zero from a parse/filter mismatch.

Both bullets are also considerably shorter than the originals; the merge/dup mechanics and the `AppendCountSanity` archaeology belong in the PR, not in release notes.

Unrelated to your two items but in the same area: `zeroCountSanityBlock` in `count_sanity.go` was not gofmt-clean and this branch introduced it. Fixed. It was the only branch-touched file gofmt flagged.

Commits are local on the branch, not pushed yet.

internal/telemetry/logs/attributes_for_pipeline.go
faiz9068: Verified fixed. Ran the real `sampleBodyForModel` at 0d4e2be:

```
IN : auth failed Authorization: Bearer eyJhbGciOiJIUzI1NiJ9.PAYLOAD.sig for user
OUT: auth failed Authorization: [redacted] for user
IN : authorization=Bearer abc123def upstream=payments
OUT: authorization=[redacted] upstream=payments
IN : Authorization: Basic dXNlcjpwYXNzd29yZA==
OUT: Authorization: [redacted]
IN : password: hunter2 user=bob        OUT: password: [redacted] user=bob
IN : basic auth failed for user bob    OUT: basic auth failed for user bob   (unchanged, no false note)
```

Bearer-first ordering plus the optional-scheme group both do what they claim, key preservation is intact, and prose containing the bare word `basic` is left byte-identical. Sharing `CredentialKeyAlternation` is the right call — and the `upstreamBodySecretPattern` rewrite is behaviour-neutral, since the alternation is character-identical to the old inline one and the capture group was already there and unused by the literal replacement.

One residual of the same class, raised separately below — the scheme allowlist is `bearer|basic`, so other auth schemes still slip through with a `credential-redacted` note attached. Not blocking. Resolving this thread.

internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -56,8 +70,137 @@ const (
...
+// separate, purpose-built copy for sampleBodyForModel: unlike the upstream
+// error-body sanitizer, this transform must preserve byte fidelity (no
+// whitespace collapsing) and preserve credential KEYS (redact only the
+// value), so it cannot share an implementation with SanitizeUpstreamBody.
+var (
+	sampleBodyURLPattern    = regexp.MustCompile(`(?i)\b[a-z][a-z0-9+.-]*://[^\s"'<>,}\]]+`)
+	sampleBodyBearerPattern = regexp.MustCompile(`(?i)\b(bearer)\s+[^\s"',}]+`)
+	// Optional scheme absorbs a Bearer/Basic prefix: without it the value class
+	// stops at the space and redacts only the scheme word, leaking the token.
+	sampleBodyCredentialPattern = regexp.MustCompile(`(?i)\b(` + utils.CredentialKeyAlternation + `)\b("?\s*[:=]\s*"?)(?:(?:bearer|basic)\s+)?[^"',}\s]+`)
```
faiz9068, line 90: Residual of the same class, non-blocking: the scheme allowlist is `bearer|basic`, so any other auth scheme leaks — and still gets a `credential-redacted` note, which is the part worth fixing.

Run at head:

```
IN : Authorization: DPoP eyJhbGciOi.TOKEN.sig
OUT: Authorization: [redacted] eyJhbGciOi.TOKEN.sig   notes: [credential-redacted]
```

Same for `Digest`, `Negotiate`, `NTLM`, `AWS4-HMAC-SHA256`. Rarer than Bearer in a log body, so severity is low — but it is the identical failure mode: value class stops at the space, scheme word absorbs the redaction, note claims a redaction that didn't happen.

Worth resisting the obvious fix of a generic `(?:\w+\s+)?` scheme group — that over-redacts, e.g. `password: hunter2 user=bob` would treat `hunter2` as the scheme and redact `user=bob` instead. Lower-risk options are extending the allowlist to the registered HTTP auth schemes, or special-casing `authorization` keys to redact to the end of the value region.

Separately: bare `Basic <b64>` with no preceding key is still untouched (`Basic dXNlcjpwYXNzd29yZA== sent` passes through verbatim, no note), because only `sampleBodyBearerPattern` covers the keyless form. Pre-existing in `SanitizeUpstreamBody` too, so not a regression — just slightly odd now that the keyed pattern knows about `basic` and the keyless one doesn't.

faiz9068 (commented):
Re-reviewed the delta (`76b1254..0d4e2be`). Both blocking items are fixed, and I verified each rather than reading for it.

**Resolved — credential ordering.** Ran the real `sampleBodyForModel` at head: `Authorization: Bearer <jwt>`, `authorization=Bearer abc123`, and `Authorization: Basic <b64>` all come out `[redacted]` with the key preserved, and prose containing a bare `basic` is left byte-identical. Details in the resolved thread. Sharing `CredentialKeyAlternation` is the right shape, and the `upstreamBodySecretPattern` rewrite is behaviour-neutral — the alternation is character-identical to the old inline one, and its capture group was already there and unused by the literal replacement.

**Resolved — changelog.** Both bullets are under `[Unreleased]` now, `sample_bodies` is named correctly, the not-PII-redacted caveat made it into the release note, the parse-or-filters-Body gate is described accurately, and `service_log_volume` is called out.

One residual of the same class inline — non-blocking. Local `go build ./...` and `go test ./internal/telemetry/logs/... ./internal/utils/...` green in a worktree at head; head-commit CI green across all six checks (`Run Tests`, CodeQL go/js/actions).

Still open from the earlier round, unchanged and still non-blocking — flagging only so they are a decision rather than an oversight: the `$eq`-on-Body gap in `bodyTouchFilterOperators`, the bare `hasParseStage` trigger firing on ordinary zeros, and the JSON/logfmt branches carrying no `sample_bodies`.

Not approving — that call is not mine to make here.

faiz9068 (approved):

PR #226, opened 2026-08-24
fix(apm): bound get_service_performance_details query cost for wide windows

prathamesh-sonpatki:
get_service_performance_details reused the full requested window as the inner PromQL range-vector selector on ~9 sub-queries (e.g. rate(x[<full window>])), which made the backend rescan the entire window's raw samples per output step. Confirmed via direct backend testing this crashes/OOMs the backend for windows wider than ~1-2 weeks, and that the backend hard-caps a single query's range at 35 days.

- Range-vector sub-queries (apdex, response times, availability, throughput, error rate, error percent) now use a fixed 5m inner selector instead of the full window.
- Windows wider than 35 days are split into consecutive <=35-day chunks, queried independently, and merged: boundary timestamps shared between adjacent chunks are deduped, and series are matched by label set rather than array position so a label combination present in only some chunks still comes through.
- The three top-k queries keep their chunk's own width as the inner selector (their only windowing mechanism); when chunked, they over-fetch more candidates per chunk and merge across chunks (max per key for the p95-latency-style query, sum per key for the two count-style queries) before truncating to the top 10.
- A window over 366 days is rejected before any HTTP call - a call-count/latency guard (each chunk costs ~9 backend calls), not a data-retention cutoff; data outside retention already comes back empty per-chunk with no error.
- Windows <=35 days are unaffected: same single query per metric, same output shape as before.

---
<a href="https://app.blacksmith.sh/last9/codesmith/last9-mcp-server/pr/226"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790163833&installation_model_id=433497&pr_number=226&repository=last9%2Flast9-mcp-server&return_to=https%3A%2F%2Fgithub.com%2Flast9%2Flast9-mcp-server%2Fpull%2F226&signature=7ae15fcf46747a5ea490743542e4055e724badb5847faf0295b0b973266bb3eb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


internal/apm/apm.go
```diff
@@ -257,279 +464,253 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
...
+		// instant queries whose only windowing mechanism is the inner
+		// quantile_over_time([...]) selector, so (unlike the rate() queries
+		// above) it must stay at the chunk's own width, not perfDetailsStepWindow.
+		topRTLimit := 10
+		if multiChunk {
+			topRTLimit = 20 // over-fetch per chunk so the cross-chunk merge still has 10 real candidates
+		}
+		var topRTChunks [][]map[string]float64
+		for _, c := range chunks {
+			chunkWindow := fmt.Sprintf("%dm", int((c.end-c.start)/60))
```
faiz9068: `int((c.end-c.start)/60)` truncates, so a trailing chunk shorter than 60s renders `[0m]` — a PromQL parse error, so that chunk 422s into `partial_errors`. Reproduced with a window of 35d+30s:

```
SELECTOR: [50400m]
SELECTOR: [0m]
```

Reachable any time an explicit start/end span exceeds a 35d multiple by <60s, and it's the most recent chunk that's lost. Clamp to a 1m minimum or format seconds (`%ds`). Same line at 613 and 668.

internal/apm/apm.go
```diff
@@ -238,6 +267,184 @@ type ServicePerformanceDetails struct {
...
+			existing, ok := merged[key]
+			if !ok {
+				cp := TimeSeries{Metric: s.Metric, Values: append([]TimeSeriesPoint{}, s.Values...)}
+				merged[key] = &cp
+				order = append(order, key)
+				continue
+			}
+			vals := s.Values
+			if len(existing.Values) > 0 && len(vals) > 0 &&
+				existing.Values[len(existing.Values)-1].Timestamp == vals[0].Timestamp {
```
faiz9068: This dedup needs exact timestamp equality, but chunks aren't equal width — a 90d window sends 35d/35d/20d, and the step is derived server-side from `window` (`MakePromRangeAPIQuery` sends no `step`). So the last chunk comes back on a finer grid, this equality won't hold, and you get near-duplicate points at the seam plus a resolution change mid-series that a consumer reads as a change in the data. Every test hand-crafts aligned boundaries, so the suite can't see it.

Two fixes, ideally both: split into `ceil(window/35d)` equal-width chunks so every chunk gets the same server-side step, and drop any incoming point with `Timestamp <= existing last` instead of matching one exact value.

internal/apm/apm.go
```diff
@@ -238,6 +267,184 @@ type ServicePerformanceDetails struct {
...
+	}
+	return result
+}
+
+// fetchChunkedRangeSeries runs a range-vector query once per chunk (via
+// buildQuery) and merges the results with mergeChunkedSeries. A chunk that
+// returns a non-2xx status is recorded in partialErrors (with its time
+// bounds) and skipped, so the rest keep merging; a transport-level error
+// aborts the whole call, matching the pre-chunking behavior.
+func fetchChunkedRangeSeries(ctx context.Context, client *http.Client, cfg models.Config, chunks []perfDetailsChunk, buildQuery func(perfDetailsChunk) string, label string, partialErrors *[]string) ([]TimeSeries, error) {
```
faiz9068: Chunks run strictly sequentially with no per-chunk timeout — at the 366d ceiling that's 11 chunks x ~9 sub-queries = ~99 serial calls, which isn't much of a latency bound, and one hung chunk stalls the whole call.

`utils.TimeChunk`, `utils.RunChunksParallel`, and `constants.PerChunkHTTPTimeout` already exist and are used by `get_logs` / `get_service_logs` / `get_traces`. Reusing them gets bounded parallelism plus per-chunk timeouts and drops the bespoke `perfDetailsChunk`.

internal/apm/apm.go
```diff
@@ -238,6 +267,184 @@ type ServicePerformanceDetails struct {
...
+		}
+		func() {
+			defer httpResp.Body.Close()
+			if httpResp.StatusCode != http.StatusOK {
+				*partialErrors = append(*partialErrors, fmt.Sprintf("%s: %s", chunkBoundsLabel(c), promErr(httpResp, label).Error()))
+				return
+			}
+			data, err := io.ReadAll(httpResp.Body)
+			if err != nil {
+				*partialErrors = append(*partialErrors, fmt.Sprintf("%s: failed to read response body for %s: %v", chunkBoundsLabel(c), label, err))
```
faiz9068: Read/parse failures used to abort the call; they're partial errors now — including for <=35d windows, which the PR says are unaffected.

It's also the opposite of the 0.15.1 decision for logs/traces (CHANGELOG: a failed later time-chunk is a tool error, the `partial_result` merge-and-continue envelope is gone). Either policy is defensible — worth making the divergence deliberate and saying so in the CHANGELOG, since a failed middle chunk here yields a series with a 35d hole in it.

internal/apm/apm.go
```diff
@@ -238,6 +267,184 @@ type ServicePerformanceDetails struct {
...
+	var chunkResults [][]TimeSeries
+	for _, c := range chunks {
+		httpResp, err := utils.MakePromRangeAPIQuery(ctx, client, buildQuery(c), c.start, c.end, cfg)
+		if err != nil {
+			return nil, err
+		}
+		func() {
+			defer httpResp.Body.Close()
+			if httpResp.StatusCode != http.StatusOK {
+				*partialErrors = append(*partialErrors, fmt.Sprintf("%s: %s", chunkBoundsLabel(c), promErr(httpResp, label).Error()))
```
faiz9068: `chunkBoundsLabel` is prefixed unconditionally, so a 1h window that gets a 500 now returns `chunk <ts>..<ts>: <msg>` instead of `<msg>`. Gate on `multiChunk` to keep the <=35d output identical as promised.

internal/apm/apm.go
```diff
@@ -257,279 +464,253 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
...
-						val, _ := strconv.ParseInt(valStr, 10, 64)
-						op := make(map[string]int64)
-						op[key] = val
-						details.TopErrors = append(details.TopErrors, op)
+		if multiChunk {
+			if len(topErrChunks) > 0 {
+				details.TopOperations.ByErrorRate = mergeTopInt64(topErrChunks, 10)
+			}
+		} else if len(topErrChunks) > 0 {
+			details.TopOperations.ByErrorRate = topErrChunks[0]
```
faiz9068: `topErrQuery` / `topErrorsQuery` aren't `topk`-wrapped, so this single-chunk path returns every row while the multi-chunk path truncates to 10. A 30d query can return 200 entries and a 40d query 10, with nothing in the response signalling truncation. Pick one — truncating both matches the field name, but it's a shape change for narrow windows and needs a CHANGELOG note. Same at 713.

internal/apm/apm.go
```diff
@@ -257,279 +464,253 @@ func NewServicePerformanceDetailsHandler(client *http.Client, cfg models.Config)
 			return nil, nil, fmt.Errorf("service_name is required")
 		}
 
-		timeRange := fmt.Sprintf("%dm", int((endTimeParam-startTimeParam)/60))
+		windowSeconds := endTimeParam - startTimeParam
+		if windowSeconds > int64(maxServicePerformanceWindowDays)*24*3600 {
+			return nil, nil, fmt.Errorf(
+				"time range too wide for get_service_performance_details: got %d days, max is %d days (this call fans out to multiple sub-queries per %d-day chunk; this bound limits call count/latency, not data availability)",
```
faiz9068, line 678: Integer division makes the boundary message contradict itself — at 366d+1s this reads `got 366 days, max is 366 days`. Round the reported figure up, or report hours.

internal/apm/apm.go
```diff
@@ -16,6 +18,33 @@ import (
...
+	// this bounds how many chunks (and how long) a single tool call can take.
+	maxServicePerformanceWindowDays = 366
+	// perfDetailsMaxChunkDays is the backend's confirmed hard per-query range
+	// cap. Windows wider than this are split into consecutive chunks of at
+	// most this many days.
+	perfDetailsMaxChunkDays = 35
+	// perfDetailsStepWindow is the fixed inner range-vector selector duration
+	// used for rate()/step-based sub-queries, replacing the old (and unsafe)
+	// "reuse the whole requested window" behavior.
+	perfDetailsStepWindow = "5m"
```
faiz9068: Not the step — the step is server-derived. `perfDetailsInnerRange` (or `...RangeSelector`) would stop a future reader assuming this controls output resolution.

faiz9068 (commented):
Core fix looks right: fixed inner selector, <=35d chunking, label-keyed merge, and max-vs-sum merge per metric type. Two things I doubted both check out — `topk(20)` is provably more than enough for a max-merge (a key outside a chunk's top-10 can't be in the global top-10), and the per-iteration `defer` restructuring is correct.

Few things below. The `[0m]` one I reproduced; the chunk-width/step one is invisible to the tests but not to production.

Also missing: a CHANGELOG entry (three user-visible changes here — the 366d hard error, wide-window values changing meaning, and top-list truncation at >35d), and the 366d cap isn't in `get_service_performance_details.md`, so a model will burn a call finding it.

One question worth settling before merge: the PR shows a 5m inner selector is *safe* at 35d, not that the output is still meaningful. With a server-derived step likely much larger than 5m at that width, `rate(x[5m])` samples a small fraction of the window, and `sum by (quantile)(trace_service_response_time{...}[5m])` returns gaps wherever there was no sample in the preceding 5m — where the old full-window selector always had something. Worth eyeballing a wide window against a low-traffic service; if you already did, ignore this.

faiz9068 (commented):
Re-reviewed `626b5cb` (delta since my last review). All 8 previous findings are addressed, and I verified the four testable ones by execution at the PR head:

- **`[0m]` selector** — gone. `chunkWindowSelector` clamps to 1m, and the equal-width split means the sub-60s remainder chunk can't occur anymore. 35d+30s now yields `[25200m 25200m]`.
- **Equal-width chunks** — confirmed: 90d → 3 × `43200m`, 366d → 11 × `47912m`, all within 1s of each other, none over 35d.
- **Guard message** — `got 367 days, max is 366 days` at 366d+1s.
- **Parallelism** — `RunChunksParallel` usage is correct: results are index-ordered, each goroutine writes its own slot, and `partialErrors` is appended in a sequential loop after `wg.Wait()`, so no race. Per-chunk `PerChunkHTTPTimeout` is applied.

Also resolved: `<=` boundary dedup robust to misaligned grids (with a test), `perfDetailsInnerRange` rename plus a comment saying what it isn't, `top_n` applied uniformly so the truncation asymmetry is gone, CHANGELOG entry, and the doc update.

The generic `fetchChunkedTopK` collapsing the three near-identical top-k loops is a nice extra — it also closes the swallowed-decode-error hole I'd raised as minor.

---

**One new issue, and it's the same class as the finding it came from.**

My earlier comment was narrow: *read/parse* failures had silently gone from hard errors to partial errors. The fix went wider and made **every** single-chunk failure hard-abort — including non-2xx, which was always soft.

`fetchChunkedRangeSeries` says this "matches the pre-chunking behavior exactly", and the CHANGELOG repeats it. That isn't right. Verified by running the same scenario on both revisions — one sub-query (throughput) returning 500 on a 1h window:

```
a16ca34 (pre-PR):  SUCCESS
  {"service_name":"svc", ... ,"partial_errors":["service performance details throughput failed with HTTP 500 ..."]}

626b5cb (head):    HARD ERROR, whole call aborted
  service performance details throughput failed with HTTP 500. The upstream service is temporarily unavailable; retry later.
```

The old contract was encoded in the test that this commit inverted — `TestServicePerformanceDetailsPromFailureRecordsPartialErrors`, whose assertion read *"HTTP PromQL failures should stay in partial_errors"*. Pre-change, only read/parse errors were hard; non-2xx was soft.

Why it matters: the ≤35d path is the common one, and it fans out to 9 sub-queries. Any single one 500ing now fails the entire call, where before you got the other 8 metrics plus a note. That's a real availability regression on the hot path, and it's strictly worse than what I asked for.

Suggested: keep non-2xx soft (`partial_errors`) on the single-chunk path as it always was, and hard-abort only on read/parse failures — that's the actual pre-change contract. If you'd rather make it uniformly hard, that's defensible too, but it should be described as a deliberate change rather than as restoring prior behavior, and both the code comment and the CHANGELOG line need correcting.

---

**Minor, non-blocking:**

- `top_n` is unbounded — `top_n: 1000000` on a wide window sends `topk(2000000, ...)` to the backend (verified). Worth a sanity cap given the point of this PR is bounding query cost.
- `chunkWindowSelector`'s 1m clamp has no test. It's unreachable via `splitIntoPerfDetailsChunks` today, so a future change to the splitter could regress it silently.
- `mergeChunkedSeries` now drops non-monotonic points inside the *first* chunk too, so it applies to ≤35d responses as well. Almost certainly harmless, just noting it's a narrow-window behavior change that isn't in the CHANGELOG.

CI is green on `626b5cb` (13 checks) and the branch is mergeable. Not approving — that's Faiz's call — but from my side the only thing I'd want changed before merge is the non-2xx regression.

internal/apm/apm.go
faiz9068: Fixed. `chunkWindowSelector` clamps to a 1m minimum, and the equal-width split removes the narrow-remainder chunk entirely. Verified at head: 35d+30s now gives `[25200m 25200m]`, no `[0m]`.

internal/apm/apm.go
faiz9068: Fixed both ways — equal-width chunks so the grids match, and a `<=` filter so the merge is robust even when they do not. Verified: 90d -> 3 x 43200m, 366d -> 11 x 47912m, all within 1s and none over 35d. `TestMergeChunkedSeries_MisalignedGridBoundary_NoDuplicatesMonotonic` covers the misaligned case.

internal/apm/apm.go
faiz9068: Fixed — now `utils.RunChunksParallel` with `perfDetailsMaxConcurrency` and `constants.PerChunkHTTPTimeout` per chunk. Checked the usage: results are index-ordered, each goroutine writes its own slot, and `partialErrors` is appended sequentially after `wg.Wait()`, so no race.

internal/apm/apm.go
faiz9068: Fixed — the chunk-bounds prefix only appears on the multi-chunk path now, and the single-chunk test asserts it is absent.

internal/apm/apm.go
faiz9068: Fixed — `mergeTopInt64` now truncates to `topN` on both paths, so narrow and wide windows behave the same. The new `top_n` arg is a good addition. One nit in the main review: it is unbounded, so `top_n: 1000000` sends `topk(2000000, ...)`.

internal/apm/apm.go
faiz9068: Fixed — ceiling division. Verified at head: 366d+1s now reports `got 367 days, max is 366 days`.

internal/apm/apm.go
faiz9068: Fixed — renamed to `perfDetailsInnerRange`, and the comment now explicitly says it does not control output resolution. Thanks.

internal/apm/apm.go
faiz9068: Half addressed, so leaving this open. The divergence is now deliberate and in the CHANGELOG, which is what I asked for. But the fix went wider than the finding: non-2xx on a single-chunk window also became a hard abort, and that one was always soft. The code comment and CHANGELOG both say this "matches the pre-chunking behavior exactly" — it does not. Verified on both revisions, and the inverted test `TestServicePerformanceDetailsPromFailureRecordsPartialErrors` asserted the old contract in its name. Details in the main review.

faiz9068 (approved):

PR #227, opened 2026-08-24
fix: teach safe query semantics

prathamesh-sonpatki:
## Summary

Agents can build safe day-wise percentiles from raw log values instead of declaring quantiles unsupported, averaging percentile outputs, or redirecting a logs-only request to traces.

- Validates `$quantile` as `[level, field]` across log and trace aggregate pipelines before execution.
- Requires parsed log attribute/resource percentile fields to pass a canonical anchored numeric regex before aggregation, and requires answers to disclose excluded non-numeric values.
- Documents one full-range `window_aggregate`, requested timezone preservation, normalized-route discovery, raw-URI non-recombination, source-unit preservation, generic count windows, filter operators, and partial-result disclosure.
- Caps aggregate log responses, probes one extra output row, and returns `l9_result.partial=true` when the row limit is reached.
- Keeps Body word searches efficient: ALL words use `$and`, ANY words use `$or`, and neither uses `$icontainsWords`.

## Validation

- Credential-free local suite: `TEST_REFRESH_TOKEN= LAST9_REFRESH_TOKEN= LAST9_DELETE_TOKEN= go test ./... -count=1` — passed.
- `go build .` — passed.
- `go test ./internal/prompts ./internal/telemetry/logs ./internal/telemetry/traces -count=1` — passed.
- `go vet ./...` — passed.
- A separate live-enabled full-suite repeat reached two unrelated dashboard CRUD tests and received 403 from the default read-only token; no dashboard code changed.
- `git diff --check` — clean.
- Added-line privacy scan — no tenant identifiers, service names, secrets, or ticket IDs.
- Companion eval unit tests: 33/33 focused tests passed; full suite 100/101, with one unrelated pre-existing APM fixture mismatch.
- Focused served-description candidate-vs-main eval (three trials each): candidate 8/9 versus `origin/main` 0/9 across log pipeline, log-only routing, and trace percentile cases. The one candidate miss stopped after field discovery; an immediate rerun was 3/3, making that routing case 5/6 overall.
- Full served-description corpus: 199/209 passed, 10 failed, 8 skipped. The focused log percentile, trace percentile, and log-only routing cases passed. Remaining failures were model-generated query/routing issues, including one broad percentile case that lost service scope, invalid filter/aggregate shapes, parse-before-group ordering, and log-tool selection disagreements.
- Dedicated agentic discovery: 2/4 passed. Scoped discovery paths were model-sensitive; the broad percentile case exhausted retries without numeric aggregate evidence.
- Parameter-mismatch suite: 17/17 passed, including all deterministic preflights, model repair cases, and the intentional negative control.
- Workflow comparisons: exception RCA 2/2 baseline and 2/2 treatment; scoped log discovery 1/3 baseline and 2/3 treatment with one fixed/no regressions; latency, error-rate, slow-query, and on-call workflows all passed in both arms, including the latency prompt-error check.
- Dedicated APM deviations: baseline 1/7 versus treatment 7/7, with one unavailable fixture skipped in both arms (+85.7 percentage points).
- Real-data replay reached the log backend without the former unsupported-quantile claim. The ambiguous workload did not expose a service-scoped endpoint/latency mapping, so no P99 was claimed. Repeated speculative probes in that no-mapping path remain a supervisor-policy issue, not an MCP quantile capability claim.

## Post-deploy validation

- Monitor log-query 4xx/5xx rates and aggregate latency for 24 hours.
- Healthy behavior: raw-value P99 uses `$quantile`; unsafe parsed numeric fields fail locally; capped results disclose partial coverage.
- Failure signals: unexpected trace/metrics routing for logs-only requests, repeated day-by-day calls, or inferred units.


internal/telemetry/logs/logs.go
```diff
@@ -97,20 +100,30 @@ func fetchLogJSONQuery(ctx context.Context, client *http.Client, cfg models.Conf
...
 				"[chunking] get_logs chunking disabled start_ms=%d end_ms=%d limit=%d index=%q",
 				startTime,
 				endTime,
-				args.Limit,
+				queryLimit,
 				args.Index,
 			)
 		}
-		result, err := executeLogJSONQuery(ctx, client, cfg, logjsonQuery, startTime, endTime, args.Limit, args.Index)
+		result, err := executeLogJSONQuery(ctx, client, cfg, logjsonQuery, startTime, endTime, queryLimit, args.Index)
```
faiz9068, line 118: Blocking question: this newly sends `limit` on aggregate pipelines. Previously `args.Limit` was typically 0, and `service_logs.go` gates on `if limit > 0`, so no `limit` param was sent at all — now it's 1001.

Is that limit applied before or after aggregation on the query-range endpoint? If before, a p99 over a busy service is computed from 1001 rows instead of the full window — silently wrong, and `l9_result.partial` wouldn't fire because the response legitimately has few aggregated rows. The tests use canned response bodies, so they can't tell the two apart.

If it's pre-aggregation, the cap should be client-side only: drop the `+1` probe and the query param, keep `applyGetLogsAggregateRowLimit`.

internal/telemetry/traces/input_schema.go
```diff
@@ -117,13 +117,13 @@ func aggregateStageSchema() map[string]interface{} {
...
 		"type":        "object",
-		"description": "Window aggregate stage: time-bucketed aggregation. Use for rates or counts over time windows.",
+		"description": "Window aggregate stage: time-bucketed aggregation. Use for rates, counts, or percentiles over time windows.",
 		"required":    []string{"type", "function", "as", "window"},
 		"properties": map[string]interface{}{
 			"type": stringEnum("window_aggregate"),
 			"function": map[string]interface{}{
 				"type":        "object",
-				"description": "Aggregation function object. Example: {\"$count\": []}",
+				"description": "Aggregation function object. Examples: {\"$count\": []}, {\"$quantile\": [0.99, \"Duration\"]}",
```
faiz9068, line 126: This teaches `$quantile` on traces, but there's no arity/order/bounds check on the traces side — a swapped `{"$quantile": ["Duration", 0.99]}` gets an opaque 400 instead of the `[level, field]` hint that `get_logs` returns.

The numeric-gate half correctly doesn't apply here (`Duration` is already numeric), but the arity/order/bounds half is domain-independent. Worth lifting `validateQuantileFunction` into a shared helper both sides call.

internal/telemetry/logs/query_validate.go
```diff
@@ -230,6 +235,164 @@ func validateLogJSONQuery(stages []map[string]interface{}, pathPrefix string) er
...
+			}
+		}
+	case []interface{}:
+		for _, item := range typed {
+			collectMandatoryCanonicalNumericRegexFields(item, fields)
+		}
+	}
+}
+
+func isCanonicalNumericRegex(pattern string) bool {
```
faiz9068, line 379: This gate drops rows: scientific notation, thousands separators and unit suffixes (`120ms`) are excluded, and the p99 is then reported over the survivors with no indication of how many were dropped.

Given `l9_result.partial` exists precisely to make truncation visible, this is inconsistent. Minimum: a line in `logjson.md` telling the model to state that non-numeric values were excluded. Better: surface the excluded-row count.

internal/prompts/references/logjson.md
```diff
@@ -237,17 +237,37 @@ Note that regex parsing operators also work as regex filters
...
-  "groupby": {"field": "alias"} // optional group-by fields
-}
+[
+  {
+    "type": "filter",
+    "query": {"$and": [{"$neq": ["ServiceName", ""]}]}
+  },
+  {
+    "type": "parse",
+    "parser": "json",
```
faiz9068, line 257: The generic `window_aggregate` shape template with `{"$count": []}` was replaced here (and in `tracejson.md`) by the percentile pipeline, and the `$count` example also went from the served description. Per-minute count trends are probably the most common `window_aggregate` use, and there's now no minimal `$count` template in either place.

The description is bounded by `TestWhaleDescriptionsBounded`, but the reference isn't — worth keeping the generic shape block here alongside the new example. Same for the filter operator list dropped from the description, since `$regex` is now required by the percentile path.

faiz9068 (commented):
Reviewed at 659768c. Built, vetted and ran `go test ./... -count=1` locally — all green, matching the description.

The validation work is solid: running the numeric-guard pass after sanitization, recursing only into `$and` (so a regex under `$or`/`$not` can't be mistaken for a mandatory guard), and modelling a later `parse` re-dirtying an already-gated field are all the right calls, and the tests attack them properly. `$quantile_exact` is untouched, so the "no new unbounded exact-log aggregation" claim checks out.

One blocker plus a few follow-ups inline. Separately: no CHANGELOG entry — this ships three observable contract changes (the `l9_result.partial` key, the new aggregate row cap, and the new hard rejections at the log query boundary), so it's worth a bullet at the top of `### Fixed`.

Worth noting the last commit adds a query-cost bound, a new response key and a dataflow analyzer — none mentioned in the PR description. Splitting that strand out would have made the limit question below easier to catch.

internal/telemetry/logs/logs.go
prathamesh-sonpatki: > Is that limit applied before or after aggregation on the query-range endpoint?

Verified: it is applied after aggregation. The log handler parses the query `limit`, and `sqlquerygen.TranslatePipeline` appends a final `select` carrying that limit after the `aggregate`/`window_aggregate` processor. The 1,001 probe therefore caps aggregated output rows, not raw log samples. Omitting it would restore the backend default of 1,000 and prevent reliable `+1` truncation detection, so no code change is needed here.

internal/telemetry/traces/input_schema.go
prathamesh-sonpatki: > Worth lifting `validateQuantileFunction` into a shared helper both sides call.

Fixed in e57a187. Log and trace aggregate paths now use one domain-independent `$quantile` argument validator. Trace `aggregate` and `window_aggregate` calls return the actionable `[level, field]` path locally for swapped, out-of-range, and wrong-arity arguments; handler tests verify invalid pipelines make zero upstream requests.

internal/telemetry/logs/query_validate.go
prathamesh-sonpatki: > Minimum: a line in `logjson.md` telling the model to state that non-numeric values were excluded. Better: surface the excluded-row count.

Addressed in e57a187 with the minimum reliable contract: the served description plus log and trace references now state that the canonical numeric regex excludes non-matching values from percentile calculations and require that exclusion to be disclosed in the answer. I did not add an excluded-row count because the aggregate response does not expose one; deriving it would require another query and a broader response-contract change.

internal/prompts/references/logjson.md
prathamesh-sonpatki: > Worth keeping the generic shape block here alongside the new example. Same for the filter operator list dropped from the description.

Fixed in e57a187. Generic `$count` `window_aggregate` examples are restored in the served log description and both log/trace references, alongside the percentile examples. The served description also restores compact logical/comparison/word/regex operator guidance and remains within its 2,600-character bound.

prathamesh-sonpatki:
> no CHANGELOG entry — this ships three observable contract changes

Added under the top Unreleased `Fixed` section in e57a187, covering capped aggregate log results with `l9_result.partial`, local quantile validation, and the numeric-field safety check.

> the last commit adds a query-cost bound, a new response key and a dataflow analyzer — none mentioned in the PR description

Updated the PR description to call out the aggregate output-row cap and probe, `l9_result.partial`, log numeric-field dataflow validation, and the new shared log/trace quantile argument validation. The validation section now also reflects the final focused tests, full-suite result, vet, and the unrelated live dashboard 403 seen on a repeat run.

faiz9068 (approved):

PR #228, opened 2026-08-25
feat(logs): answer get_logs and get_service_logs with a single server-side search call, behind a flag

faiz9068:
## What

`get_logs` and `get_service_logs` currently answer a search by slicing the time range into chunks client-side and fanning out one upstream request per chunk. This adds a second path that hands the whole search to the API in one call — the server plans, probes, and fetches — and puts it behind a flag.

`LAST9_USE_LOG_SEARCH_API=true` (or `--use_log_search_api`) enables it. **Off by default**, so this PR changes nothing until someone opts in.

## Why

Chunk fan-out costs one upstream request per hour of window. On a 14-day search that's hundreds of requests, and any one of them failing fails the whole call — or worse, returns a partial answer that looks like a complete one.

## Output shape

Unchanged on both tools, plus new sibling keys.

`get_logs`, on raw searches:

- `total_matching_lines` — counts all matching lines, not the returned sample. A floor rather than a total when `logs_truncated` is true or `search_stats.chunks_failed` is above zero.
- `logs_truncated`
- `search_stats` — chunks planned / with data / failed, bucket width, covered range
- `volume_summary` — densest buckets, to narrow a follow-up search

Aggregate queries get `search_stats` only.

`get_service_logs` gains `total_matching_lines` and `search_stats`, both omitted when the server does not report them, so the chunked path's output is byte-identical to today's.

## Structure

The new path is self-contained in `internal/utils/log_search_api.go`, reached through a single branch at the top of `fetchLogJSONQuery` and `fetchServiceLogs`. `get_service_logs` reuses its existing response parser unchanged, since `query_result` is lifted to where that parser already reads. The chunking code is untouched. Once this is proven in production, deleting the old path is removing those branches and the chunking file — no untangling.

## Testing

Verified end-to-end against two production orgs, comparing both paths against a **direct API call as an independent oracle** rather than against each other:

- **Correctness** — every case exact: same line set, zero missing, zero extra, ordering preserved, identical aggregate group counts. Complete-set reassembly verified at 14-day windows (2300/2300, 1961/1961, 2403/2403 lines).
- **Performance** — 2.2–9.3× faster on wide windows (~5.3× at 72h, ~5.5× at 7d, ~4.9–8.6× at 14d). Roughly a wash at ≤6h. Aggregates and body-text search are scan-bound, so neutral.
- **Agent evals** — tool surface byte-identical (44 tools). Across four agentic suites, 23/29 on chunking vs 26/29 on the new path; all three differing cases re-ran as noise.

The end-to-end runs above covered `get_logs`. `get_service_logs` has unit coverage on the new path but has not yet been exercised against live data.

Unit tests in `internal/utils/log_search_api_test.go`, `internal/telemetry/logs/log_search_api_test.go`, and `internal/telemetry/logs/service_logs_search_api_test.go`.

## Review follow-ups

- Rejects a response whose `query_result` is absent or null instead of returning an empty envelope or panicking. Today's endpoint sends neither; this is hardening against a future path that forgets the field.
- Covers the aggregate row cap and its interaction with the count-sanity block on the new path.


prathamesh-sonpatki (changes requested):
Re-reviewed the delta from `813cc05` to `05c2a80`.

Resolved in this update:
- `get_service_logs` now uses one server-side search call behind the flag.
- Aggregate row-cap / count-sanity composition now has coverage.
- Missing, null, and non-object `query_result` values now fail.

Verification:
- `go test ./...` passes.
- `go test -race ./internal/utils ./internal/telemetry/logs` passes.
- All 13 head check-runs are complete (success or intentionally skipped).
- Remote branch and reviewed head both resolve to `05c2a801f3ced98fd538929a7d997d4d4b161598`.

Changes still requested:

1. **P1 — ENG-1689's completeness contract remains absent.** `internal/utils/log_search_api.go:117` still exposes only the existing advisory fields, not `range_complete`, `result_complete`, searched/failed intervals, termination reason, duration, and warnings. Missing or malformed advisory fields are silently omitted, so an empty partial result can still look definitive.

2. **P1 — the service path drops the signal that makes its count a floor.** `internal/telemetry/logs/service_logs.go:423` forwards `total_matching_lines` and `search_stats`, but not `logs_truncated`. I executed the exact case `logs_truncated=true`, `chunks_failed=0`: the MCP payload returned `total_matching_lines: 41` and no truncation signal. Preserve the signal (or omit/relabel the count when incomplete), then document the result semantics in the served short description/resource so the model can qualify it.

3. **P1 — invalid-object envelopes still become false emptiness.** `internal/utils/log_search_api.go:197` proves only that `query_result` is a JSON object. `{"query_result":{}}` still decodes successfully; `get_service_logs` then returns zero logs. Validate the required `data`, `resultType`, and `result` shape in the shared decoder.

4. **P2 — three prior boundary findings are untouched:** flagged windows above the API's 14-day hard limit still fail; exactly 10,000 dataframe rows can still look complete; and quoted `NaN`/`Inf` volume counts are accepted. I reconfirmed the last one: `"NaN"` reaches `volume_summary` and `json.Marshal` fails with `unsupported value: NaN`.

The direction is good, but the rollout still cannot distinguish complete absence from partial/invalid absence reliably, so I cannot approve this head yet.

faiz9068:
Pushed `d739c60`.

**1. Completeness contract** — out of scope for this PR, as decided. The upstream scope was to move the existing chunking logic behind a single endpoint, which is what shipped; the contract fields are not part of it, so there is nothing here to forward. They get added upstream when we need them.

**2. `logs_truncated` dropped on the service path** — correct, and it was the field that made the count meaningful. Now forwarded alongside `total_matching_lines`, with a test asserting it reaches the payload. Not documenting it in the tool description yet: the description is served in both flag states, so pointing the model at fields that are usually absent is premature. That lands when the flag goes.

**3. Invalid-object envelopes** — reproduced: `{"query_result":{}}` passed the object check and returned zero logs. The decoder now requires `data`, `data.resultType`, and the `result` key; a null `result` still passes, since that is the legitimate empty-streams shape. Presence, decode and shape checks are consolidated into one function rather than split across the call site.

**4. Boundary findings** — one fixed, two I'd push back on:

- **Quoted `NaN`/`Inf`** — reproduced exactly, including the `json.Marshal` failure. Non-finite counts now drop that series; the remaining series still summarise.
- **Windows above the upstream limit** — the design leaves that bound to the server. Capping it here duplicates a constant that will drift from the one actually enforcing it, and the upstream rejection reaches the model intact.
- **Exactly 10,000 dataframe rows looking complete** — I don't think this reproduces. The aggregate cap is `min(limit, MaxGetLogsEntries)` and `MaxGetLogsEntries` is 5000, so a 10,000-row result is always trimmed and always carries `l9_result.partial`. Tested at limits 0, 5000 and 10000 — 1000/5000/5000 rows, `partial: true` in all three.

`go vet ./...`, the full suite, and `go test -race ./internal/utils ./internal/telemetry/logs` all pass.

prathamesh-sonpatki (dismissed):
Re-reviewed `05c2a801..848ce9df` (including the late main merge).

Resolved:
- `get_service_logs` now preserves `logs_truncated` alongside `total_matching_lines`.
- Empty/missing `query_result.data`, `resultType`, and `result` keys are rejected.
- Direct `NaN`/`Inf` bucket counts are dropped.
- The main merge only updates `package.json` to 0.15.2.

Verification:
- `go test ./...` passes on `848ce9df`.
- `go test -race ./internal/utils ./internal/telemetry/logs` passes on the behavior-changing parent; the late merge does not touch Go code.
- All 13 current-head checks are complete (success or intentionally skipped).
- PR head and remote branch both resolve to `848ce9df2591d735c2439696d17cd17cb2482c07`.

Changes still requested:

1. **P1 — ENG-1689's explicit completeness contract is still absent.** `internal/utils/log_search_api.go:118` still has no `range_complete`, `result_complete`, searched/failed intervals, termination reason, duration, or warnings. The served `get_logs` / `get_service_logs` descriptions also do not teach the model how to interpret the advisory fields. Missing or malformed coverage can therefore still make a partial empty result look definitive.

2. **P1 — result-shape validation is only partially fixed.** `internal/utils/log_search_api.go:204` checks that the `result` key exists, not that its value has the shape required by `resultType`. I executed `{"query_result":{"data":{"resultType":"streams","result":{}}}}`: it is accepted, and the service parser treats it as zero logs. Validate `result` as an array (or null only for the explicitly supported empty-streams shape) in the shared decoder.

3. **P2 — the range and dataframe saturation findings are untouched.** Flagged windows above the API's 14-day hard limit still route unconditionally to one rejected request, and exactly 10,000 dataframe rows can still look complete.

4. **P2 — the non-finite guard can be bypassed by aggregation overflow.** Each bucket count is checked before `summarizeLogVolume`, but two individually finite max-float counts at the same timestamp sum to `+Inf` at `internal/utils/log_search_api.go:295`. I executed that payload and `json.Marshal` still fails with `unsupported value: +Inf`. Reject/drop a bucket if the accumulated total becomes non-finite.

Good progress, but the same false-completeness and malformed-advisory failure classes remain, so this head is not ready for approval.

prathamesh-sonpatki (changes requested):
Correction to my prior review: I missed Faiz's scope clarification in issue comment #5436096778. I have dismissed that overbroad review.

Withdrawn as blockers for this PR:
- The explicit completeness contract and related tool-description work are outside the agreed scope.
- The upstream-owned 14-day limit is a deliberate design choice.
- The dataframe saturation concern does not reproduce with the current default `MaxGetLogsEntries=5000` cap.
- The aggregation-overflow case is theoretical hardening, not a merge blocker.

Confirmed resolved:
- `get_service_logs` preserves `logs_truncated`.
- Missing `query_result.data`, `resultType`, and `result` keys are rejected.
- Direct non-finite volume counts are dropped.

One change remains requested:

**P1 — validate `data.result` against `resultType`.** `internal/utils/log_search_api.go:204` checks only that the `result` key exists. On current head `848ce9df`, I independently ran `{"query_result":{"data":{"resultType":"streams","result":{}}}}`; `DecodeLogSearchResponse` accepted it, and this malformed result can become a false zero-log response downstream. Require an array for `streams` (and each supported non-empty result type), while retaining `null` only for the explicitly supported empty-result shape.

Verification: the current head is still `848ce9df2591d735c2439696d17cd17cb2482c07`, all 13 head checks are complete with success or intentional skips, and the exact malformed-result reproduction passes through the decoder.

faiz9068:
Pushed `ac84d30`.

**Validate `data.result` against `resultType`** — reproduced: an object and a string both passed and became zero logs downstream. `result` must now be an array, and the error names the type that arrived instead. Tests cover object, string and number, alongside the existing missing-key cases.

One deviation, flagged rather than buried: `null` is still accepted for every result type, not only `streams`. Verified live against both paths — an empty raw search and an empty aggregate both return `[]` on this endpoint, so the strict version would have passed too. But the chunked path does return a null `result` for a raw search with no matches, and aggregate queries pass the upstream body through untouched, so a null `matrix` stays possible. I'd rather not reject a shape the upstream can still emit for a case that isn't the reported bug. Happy to tighten it if you disagree — the bug you found is closed either way.

Verified end-to-end on both flag states: raw with and without matches, aggregate with and without matches, and `get_service_logs` with and without matches — all six identical across paths, no decoder rejections.

`go vet ./...`, the full suite, and `go test -race ./internal/utils ./internal/telemetry/logs` all pass.

prathamesh-sonpatki (approved):
Re-reviewed `848ce9df..ac84d303`.

Resolved:
- `decodeQueryResult` now rejects object, string, and numeric `data.result` values instead of allowing malformed responses to become false zero-log results.
- The validation remains in the shared decoder, so both `get_logs` and `get_service_logs` use it.
- Retaining `null` as a compatible empty-result shape is reasonable and does not reopen the reported bug.

Independent verification on `ac84d3034ea47e4558339c76fbb6c9fea9fe88a9`:
- Focused decoder tests pass, including the new malformed-shape cases.
- `go vet ./...` passes.
- `go test ./...` passes.
- `go test -race ./internal/utils ./internal/telemetry/logs` passes.
- All 13 head check-runs are complete with success or intentional skips.
- Head is stable and mergeable.

Approved.

PR #232, opened 2026-09-04
ENG-1823 Adaptive step and rate interval for get_service_performance_details

faiz9068:
Makes `get_service_performance_details` ask for a bounded number of points per series, and sizes each range selector from the step the server picks, so a wide window aggregates rather than spot-samples on the four `rate()`-based series.

## The problem

The 6 range sub-queries ran at a fixed 60s step with a hardcoded `[5m]` selector. Over 35 days that is 50,401 evaluation timestamps per series — a multi-megabyte response describing what a few hundred points describe just as well.

Coverage was not the issue: a 5m selector ticking every 60s sweeps the whole window with overlap. The issue is resolution. On a series with intermittent traffic most of those 50,401 points are empty, so the payload is enormous and the result reads as sparse.

## Why the step alone is not the lever

Step and range selector are independent in PromQL, and the backend never reconciles them. Widen the step past the selector and you get the same point count with the opposite correctness: clean-looking numbers computed from a fraction of the window, with nothing in the response indicating data was skipped.

The two halves are therefore not separable, and the commits land in that order — no intermediate commit has a widened step against a narrow selector.

## Note on the apdex revert (eda272f)

That commit's message says dropping the rollup wrapper leaves "identical coverage." That reasoning is wrong: `default_rollup(x[4h])` reads one sample from the bucket, `avg_over_time(x[4h])` reads all of them. Point density is the same; aggregation fidelity is not.

The revert itself stands — this is a resolution change, and preserving apdex's existing last-value semantics is the correct scope. The consequence, now stated in the changelog, is that apdex and response times get *sparser* on a wide window rather than aggregating it. The other half of that commit's reasoning — that VictoriaMetrics sizes a bare selector from the step — is correct and load-bearing.

## Testing

Unit tests assert on the outgoing request body: 5 of the 6 range queries carry `$__rate_interval` and none carries a literal duration selector; the budget reaches every range query and no top-k query; a zero resolution puts neither field on the wire. Apdex deliberately carries no selector, per the revert note above. Each builder is pinned to its own expected occurrence count rather than to a total, so a selector moving between queries is attributed rather than cancelling out.

The point budget is per chunk, not per merged series, so a window past 35 days returns proportionally more than ~200 points. Deliberate — the payload stays bounded — and now pinned by an assertion in the chunked test rather than implied by the single-chunk one.

Verified end-to-end at 1h / 3h / 24h / 7d / 35d / 90d — narrow windows unchanged, wider windows capped, the 90-day path still chunking and merging in order, no partial errors at any width. The 35-day case previously returned `422 Too many samples queried`.

The two backend behaviours this rests on were measured directly rather than assumed:

- `max_data_points` is a cap, not a target. A 1h window returns 61 points at a 60s step with and without the budget — identical. The step never drops below 60s (a 15m window gives 16 points at 60s). The cap only binds past 200 × 60s ≈ 3h20m: 109s step at 6h, 435s at 24h, 15196s at 35d.

The budget is applied **per chunk**, and `mergeChunkedSeries` concatenates, so a window wider than 35 days returns roughly 200 × chunk count points rather than ~200 total (90 days → 3 equal 30-day chunks → ~600 points). That is deliberate: the payload stays bounded, and equal-width chunks keep every chunk on one grid, which is what the boundary dedup in `mergeChunkedSeries` depends on. The changelog, the tool description and the chunked test now all state it as a per-chunk cap.
- `$__rate_interval` substitutes as `max(step + scrape, 4 × scrape)`. Measured widths: 240s at a 60s step (1h and 15m), 495s at a 435s step (24h), ~15270s at a 15196s step (35d). It is never narrower than the step, and never narrower than the scrape interval — the second floor is what keeps `rate()` on a narrow window from returning empty. `rate()` returned data at every width tested.

On a low-traffic series over 35 days, output goes from 50,401 points to 199, and the fraction of returned points that carry data rises from roughly a seventh to about two thirds. The same window is read either way; the buckets are simply wide enough to be worth plotting.

Closes ENG-1823. Also resolves ENG-1826.

### DIFF
--- a/internal/apm/apm.go
+++ b/internal/apm/apm.go
@@ -61,7 +61,8 @@ const (
 	// Caps output timestamps per series per chunk. Only safe because every
 	// rate() selector is sized from the resulting step via $__rate_interval;
 	// the apdex and response-time queries carry no rate() and just resolve as
-	// last-value at the wider step. The cap only binds past ~3h20m of window.
+	// last-value at the wider step. The cap only binds past ~3h20m of window,
+	// and applies per chunk - a chunked window merges to ~200 x chunk count.
 	perfDetailsMaxDataPoints = 200
 )
 
--- a/internal/apm/apm.go
+++ b/internal/apm/apm.go
@@ -30,10 +30,8 @@ import (
 //   - splitting a wider outer window into <=35-day chunks and querying each
 //     with a step-derived inner selector works cleanly.
 //
-// A step-derived inner selector ($__rate_interval) is safe at any outer
-// width up to that 35-day cap by construction, not by measurement: it is
-// never narrower than the step, so scan overlap stays near 1x instead of
-// the up-to-5x the old fixed-window form incurred.
+// A step-derived selector ($__rate_interval) is never narrower than the
+// step, so overlap stays near 1x — by construction, not measurement.
 const (
 	// maxServicePerformanceWindowDays is NOT a data-retention cutoff — a
 	// chunk entirely outside retention just comes back empty from the
@@ -58,10 +56,8 @@ const (
 	// backend in parallel, matching the small bounded concurrency other
 	// chunked callers (get_logs/get_traces) use.
 	perfDetailsMaxConcurrency = 5
-	// perfDetailsMaxDataPoints caps output timestamps per series per chunk.
-	// The 422 sample guard is a per-series tick limit — 50,401 ticks over 35
-	// days trips it, 200 sits far clear. Only safe because every range
-	// selector is sized from the resulting step via $__rate_interval.
+	// Caps output timestamps per series per chunk. Only safe because every
+	// range selector is sized from the resulting step via $__rate_interval.
 	perfDetailsMaxDataPoints = 200
 )
 
@@ -463,7 +459,7 @@ func fetchChunkedRangeSeries(ctx context.Context, client *http.Client, cfg model
 				defer cancel()
 			}
 
-			httpResp, err := utils.MakePromRangeAPIQueryWithResolution(
+			httpResp, err := utils.MakePromRangeAPIQuery(
 				chunkCtx, client, buildQuery(c), c.start, c.end, cfg,
 				utils.PromResolution{MaxDataPoints: perfDetailsMaxDataPoints},
 			)
@@ -1923,7 +1919,7 @@ func NewPromqlRangeQueryHandler(client *http.Client, cfg models.Config) func(con
 			return nil, nil, err
 		}
 
-		httpResp, err := utils.MakePromRangeAPIQuery(ctx, client, query, startTimeParam, endTimeParam, queryCfg)
+		httpResp, err := utils.MakePromRangeAPIQuery(ctx, client, query, startTimeParam, endTimeParam, queryCfg, utils.PromResolution{})
 		if err != nil {
 			return nil, nil, err
 		}
--- a/internal/apm/apm.go
+++ b/internal/apm/apm.go
@@ -25,12 +25,15 @@ import (
 // (e.g. rate(x[<full window>])), which makes the backend rescan the entire
 // window's raw samples per output step. That crashes/OOMs the backend for
 // windows wider than ~1-2 weeks. Confirmed via direct backend testing:
-//   - a step-derived inner selector ($__rate_interval) is safe at any outer
-//     width up to 35 days.
 //   - the backend hard-caps a single query's outer range at 35 days (HTTP
 //     422 "Too many samples queried" just past that).
 //   - splitting a wider outer window into <=35-day chunks and querying each
 //     with a step-derived inner selector works cleanly.
+//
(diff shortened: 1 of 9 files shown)


kneurgao (commented):
3 new findings:
- P2 — `max_data_points` budget applies per chunk, not per merged series (`internal/apm/apm.go:469`)
- P2 — no test asserts `max_data_points` across the multi-chunk (>35d) path (`internal/apm/service_performance_details_window_test.go:800`)
- P2 — tool description still claims sparse non-zero data is normal, now stale for wide windows (`internal/prompts/descriptions/get_service_performance_details.md:13` — file untouched by this PR, can't attach an inline comment there; noting it here instead)

internal/apm/apm.go
```diff
@@ -462,7 +463,10 @@ func fetchChunkedRangeSeries(ctx context.Context, client *http.Client, cfg model
 				defer cancel()
 			}
 
-			httpResp, err := utils.MakePromRangeAPIQuery(chunkCtx, client, buildQuery(c), c.start, c.end, cfg)
+			httpResp, err := utils.MakePromRangeAPIQuery(
+				chunkCtx, client, buildQuery(c), c.start, c.end, cfg,
+				utils.PromResolution{MaxDataPoints: perfDetailsMaxDataPoints},
+			)
```
kneurgao, line 470: **[P2]** `PromResolution{MaxDataPoints: perfDetailsMaxDataPoints}` is passed unconditionally per chunk in `fetchChunkedRangeSeries`, and `mergeChunkedSeries` concatenates each chunk's results. For a window split into N chunks (e.g. 90 days -> 3 chunks of ~35/35/20 days), the merged series ends up with roughly `200*N` points, not the ~200 total the PR body/changelog describe ("caps each series at about 200 data points"). Only the single-chunk case (<=35 days, tested here) actually hits ~200 total. Either scale the per-chunk budget down by chunk count before building the query (mirroring the existing `multiChunk`/`topRTLimit` over-fetch pattern a few lines below), or correct the changelog/PR-body wording to state the cap is per-chunk, not per merged series.

internal/apm/service_performance_details_window_test.go
```diff
@@ -798,8 +798,8 @@ func TestServicePerformanceDetails_QueryStrings_ChunkedUsesFixed5mRangeAndChunkW
 		t.Fatalf("expected 3 throughput range queries (one per chunk), got %d", len(rangeQueries))
 	}
 	for _, q := range rangeQueries {
```
kneurgao, line 803: **[P2]** This multi-chunk test (3 chunks) checks that every chunked range query carries `$__rate_interval`, but nothing here asserts `max_data_points` on those same chunked requests. `service_performance_details_resolution_test.go`'s new `max_data_points` assertions only exercise a single 10-day (single-chunk) window, so there's no coverage locking in what the actual multi-chunk point budget should be -- which would have caught the per-chunk-vs-merged-total question raised on `apm.go:469`.

internal/apm/apm.go
faiz9068: Confirmed — the budget is passed per chunk and `mergeChunkedSeries` concatenates, so a >35d window returns ~200 x chunk count.

Taking the wording option rather than scaling the budget: this is a resolution-only change, and scaling down would make a 90-day view coarser than a 35-day one is today. The changelog and tool description no longer imply a merged total, and d927708 pins the per-chunk value in the chunked test so switching to a total fails loudly instead of silently contradicting the docs.

One correction on the detail: 90 days is not 35+35+20. `splitIntoPerfDetailsChunks` equalises widths on purpose (3 x 30d here) — unequal widths come back on different grids and break the boundary dedup in `mergeChunkedSeries`. So there is no mixed-resolution seam, only the point count. The stale test comment saying 35+35+20 is fixed in the same commit.

internal/apm/service_performance_details_window_test.go
faiz9068: Added in d927708 — the chunked test now captures `max_data_points` per chunk and asserts each carries the full budget. Verified by mutation: dividing the budget by chunk count fails on all three chunks. That puts the per-chunk contract at the point where it matters rather than leaving it implied by the single-chunk test.

faiz9068:
Third finding (tool description, `get_service_performance_details.md:13`) — fixed in d927708, trimmed in 66dfe6c.

That line told a reader a low proportion of non-zero points is normal on a wide window. It was written for the old dense 60s grid; with ~200 wide buckets most of them carry data, so it now says a wide window returns wide buckets rather than a dense per-minute grid. Correct that the file was untouched by this branch — but this branch is what made it stale, so it belongs here.

All three now addressed; CI green on 66dfe6c.

kneurgao (commented):
No new issues. Both prior comments look addressed — the chunk-budget behavior is now documented in the code comment, tool description, and pinned with a new test assertion (internal/apm/service_performance_details_window_test.go); CI is green. 2 older threads still open — mind marking those resolved?

kneurgao (approved):

PR #199, opened 2026-07-29
feat: adopt MCP 2026-07-28 (go-sdk v1.7.0)

prathamesh-sonpatki:
## Summary

- bumps `modelcontextprotocol/go-sdk` to **v1.7.0** for MCP 2026-07-28 protocol support
- bumps `last9/mcp-go-sdk` to **v0.1.4** for anonymous stateless-client isolation and per-request `_meta` attribution
- uses the SDK's instrumented Streamable HTTP handler so telemetry records `mcp.server.transport=streamable`
- adds cache TTL middleware (SEP-2549) for static surfaces: `tools/list` / `prompts/list` (10m), `resources/list` / `resources/templates/list` (15m), `resources/read` (1h), and `server/discover` (10m)
- HTTP tests cover `server/discover` without `initialize`, plus a legacy `2025-11-25` `initialize` handshake

The HTTP handler remains stateless for horizontal scaling. Anonymous requests now receive isolated per-request client/query state and are cleaned up after completion; identified clients retain query correlation through `_meta`.

SDK fix and release: [last9/mcp-go-sdk#30](https://github.com/last9/mcp-go-sdk/pull/30), `v0.1.4`.

## Test plan

- [x] `go test ./...`
- [x] `go test -race .`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `TestStatelessStreamableHandler`, including `server/discover` result contract and legacy `initialize`
- [x] cache middleware unit + live-server TTL tests
- [x] `dump-tools` / `dump-prompts` identical to `origin/main` (44 tools, 6 prompts)
- [x] local HTTP server: two concurrent real `tools/call` requests returned 200 with distinct `unattributed_streamable_*` client/query IDs
- [x] anonymous sessions were removed immediately after both tool calls
- [x] identified client's second tool call continued the first query

PR #274, opened 2026-09-10
fix: identify the client behind each HTTP tool call

faiz9068:
## What this is

Two changes, and the second one is the actual fix:

1. `last9/mcp-go-sdk` v0.1.2 → **v0.1.4** (which transitively requires `modelcontextprotocol/go-sdk` v1.5.0 → **v1.7.0** — declared in that SDK's own `go.mod`, not an independent choice).
2. **`http_server.go` now builds the Streamable HTTP handler with the instrumented `h.server.NewStreamableHTTPHandler` instead of the upstream `mcp.NewStreamableHTTPHandler`.** The now-dead `newStatelessStreamableHandler` wrapper is deleted.

Everything else in the diff is a consequence: `go.sum`, one repointed test, one CHANGELOG line.

## Why — the bug

**In HTTP mode, every tool call was attributed to the same client.** Not "sometimes wrong" — one shared bucket for all of them, no matter how many different clients were connected.

The chain:

- The SDK decides which client a tool call belongs to by looking at the transport it is serving. It only trusts its cached client id when that transport is `stdio` (a single-client transport where the cache is safe); for anything else it deliberately returns a placeholder rather than risk attributing a call to the wrong client.
- That transport field is set by the SDK's own handler constructor. This server was calling the **upstream** constructor, which never touches it — so the field stayed empty in every HTTP deployment.
- Empty is not `stdio`, so every single HTTP tool call took the placeholder branch and came back as the same anonymous client.

The visible effect is in telemetry: because spans are grouped by client id, tool calls from unrelated concurrent sessions were folded into one trace. You could not tell which client asked for what, and per-client volume and latency were meaningless.

Switching to the instrumented constructor sets `mcp.server.transport=streamable`, which is what makes attribution work at all over HTTP. v0.1.4 supplies the other half: anonymous stateless requests now get an isolated per-request client id instead of colliding on a single shared one.

## Scope

This restores **per-request** attribution. It does not group a client's tool calls into agent turns — that needs a separate change to read the turn key off the incoming request, and is tracked separately. Expect one identity per call after this, which is correct and is already a large improvement on one identity for everyone.

## What this does not change

- The handler stays **stateless** (`StreamableHTTPOptions{Stateless: true}`). The multi-replica reason for that is unchanged, and its regression test still passes against the new handler.
- No tool, prompt, or resource behavior. `dump-tools` and `dump-prompts` are byte-identical to `main` (46 tools, 6 prompts).
- No `package.json` version bump.

## Relationship to #199

#199 proposed the same two changes, but bundled with an unrelated SEP-2549 cache-TTL middleware and `server/discover` documentation. That bundle sat unreviewed and went stale (68 commits behind `main`, four conflicting files). This PR is branched fresh off `main` and carries only the attribution fix; #199 is closed with the cache-TTL work called out as still worth doing on its own.

## Test plan

- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test -race ./...` — passing. Two `internal/dashboards` integration tests fail in a local run for want of credentials; verified failing identically on `main`, and green in CI.
- [x] `TestStatelessStreamableHandler` repointed at the instrumented handler: a request carrying an `Mcp-Session-Id` this instance never issued still returns 200 rather than 404, and `GET` still returns 405 — i.e. no regression to stateful mode.
- [x] `dump-tools` / `dump-prompts` diffed against `main` — identical.

Branch is named `release-*` so CI publishes a multi-arch image, to verify against a real deployment before merge.


faiz9068:
Closing without merging — not abandoned, deferred on purpose.

This PR bumps the SDK to v0.1.4, which stops HTTP tool calls from collapsing into one shared client identity. The remaining half of the work — reading the turn key off the request, setting the turn id on the span, and deleting the invented parent span and the `tools/list` turn delimiter — all lives inside `mcp-go-sdk` and needs another SDK release. Rather than land two dependency bumps and two reviews back to back, this repo will get **one** PR at the end that bumps straight to that final SDK version and switches to the instrumented handler.

The branch `release-sdk-v0.1.4` stays as-is; nothing here is lost, and it becomes the base for that single PR.

faiz9068:
Reopened. The scope question that led me to close this has been settled the other way: this PR **is** the whole of the client-attribution work, and the turn-grouping half has been split out into its own ticket with its own PRs in the SDK and the proxy.

So nothing here needs to wait on another SDK release. This is the one PR for attribution: set the transport via the instrumented handler, bump the SDK to v0.1.4. Ready for review as-is — CI is green and the image is published for deploy testing.

Turn grouping — reading the turn key off the request, tagging the span with it, and removing the invented parent span and the `tools/list` turn boundary — is deliberately **not** in here. It changes trace shape, needs a change in another repo to work at all, and has an open product question attached. Reviewing it alongside a dependency bump is what stalled the previous attempt.

prathamesh-sonpatki (approved):

PR #287, opened 2026-09-13
feat(mcp): add trusted API source catalog

prathamesh-sonpatki:
Exact telemetry lookups could produce false no-data conclusions because the agent had no bounded, completeness-qualified source inventory or executable source descriptor. This adds a strict catalog contract and structured candidate suggestions while keeping candidate ordering separate from correctness.

## Summary

- Add `get_api_source_catalog` with exact requested/effective bounds and explicit partial metadata.
- Fetch and validate Last9-managed source descriptors by datasource ID; no contract file, tenant provisioning, or MCP restart is required.
- Fetch the current managed contract again before each exact-quantile execution, resolving the configured or default datasource identity, authorizing verified fields, and failing closed on API, identity, or validation errors.
- Request exact backend bounds for log and trace schema discovery.
- Return positive field observations but fail closed for contract-dependent conclusions when the contract API or completeness attestation is unavailable.
- Add opt-in structured `did_you_mean` suggestions capped at three; suggestions never prove identity.

## Validation

- [x] Full race suite: 1,727 passed across 21 packages
- [x] Build
- [x] Strict schema, exact PR diff, and whitespace checks
- [x] No runtime contract-file references

Dependencies:
- Last9 API defaults and overrides: https://github.com/last9/last9/pull/11545
- Backend completeness metadata: https://github.com/last9/tempo-api-for-traces/pull/263
- Updated design: https://github.com/last9/ai/pull/313

Linear: https://linear.app/last9/issue/ENG-1904


internal/telemetry/catalog/catalog.go
```diff
@@ -0,0 +1,593 @@
...
+	}
+	if err != nil {
+		return nil, ResultEnvelope{}, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, ResultEnvelope{}, fmt.Errorf("inventory returned status %d", resp.StatusCode)
+	}
+	var payload map[string]any
+	if err := json.NewDecoder(resp.Body).Decode(&payload); err != nil {
```
nishantmodak, line 391: [P2] Preserve integer counts before converting the inventory response.

Decoding this response into map[string]any converts every JSON number to float64 before number() can validate it. A valid backend count of 9007199254740993 becomes 9007199254740992, and the catalog still returns partial=false. This defeats the exact-count contract even when the backend preserves its 64-bit integer.

Use a decoder with UseNumber (number already supports json.Number), or a typed/raw-message count field, and reject values outside the supported integer range. The consolidated review contains an executed handler regression using a synthetic HTTP response; it demonstrates the changed count on this head.

internal/telemetry/logs/attributes_for_pipeline.go
```diff
@@ -549,10 +565,33 @@ func sampleBodyDerivedAttributes(ctx context.Context, client *http.Client, cfg m
...
 	}
 
-	return nil
+	return nil, len(lines), partialErr
+}
+
+func bodySamplePartialError(result map[string]interface{}) error {
+	raw, present := result["l9_result"]
+	if !present {
+		return nil
```
nishantmodak, line 577: [P2] Mark an unattested body sample incomplete in the catalog path.

When the series response has partial=false but the raw-log sample omits l9_result, this branch treats the missing attestation as success. DiscoverLogAttributesForCatalogWithStatus then reports Partial=false, and a fields-only catalog with a configured Body field publishes top-level partial=false despite having no backend completeness statement for that sample. This can occur when endpoints differ in attestation support during rollout; the new series/inventory parsers already fail closed for the same missing field.

Return a missing-attestation error here so the catalog preserves positive sampled fields but marks its result partial. The compatibility array path already discards this error, so its best-effort behavior can remain intact. The attached handler regression uses the repository's existing bodySampleResponse fixture and fails on this exact branch.

nishantmodak (changes requested):
Reviewed all 30 changed files at `69a6c60f9a7c2fdcb1e88c762511b75ca1756d2e`, including descriptor validation, API fetch/lookup boundaries, catalog construction, exact-quantile authorization, query validation, tool registration, and tests.

Requesting changes for two reproduced issues: inventory counts lose integer precision before being labeled complete, and missing body-sample attestation is accepted as complete. The inline comments identify the required changes.

Validation: existing race-enabled tests pass across the root package, catalog, logs, traces, telemetry, suggest, toolsets, utils, and prompts. Current-head CI test and CodeQL checks are green. The two additional tests below fail locally. All HTTP responses in these regressions are synthetic local fixtures; no live Last9 backend or deployment was tested.

Save as `internal/telemetry/catalog/review_regression_test.go`, then run:

```sh
go test ./internal/telemetry/catalog -run '^TestReviewCatalog' -count=1 -v
```

Expected: preserve count 9007199254740993 exactly; mark the catalog partial when its Body sample has no l9_result envelope. Actual on this head: count 9007199254740992 with partial=false; unattested Body sample also returns partial=false. This attachment reuses testConfig, testContracts, and bodySampleResponse already present in this PR.

```go
package catalog

import (
	"context"
	"encoding/json"
	"fmt"
	"net/http"
	"net/http/httptest"
	"testing"

	"last9-mcp/internal/constants"
	"last9-mcp/internal/utils"
)

func TestReviewCatalogPreservesExactInventoryCounts(t *testing.T) {
	for _, count := range []int64{9007199254740993} {
		t.Run(fmt.Sprint(count), func(t *testing.T) {
			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
				if r.URL.Path == constants.EndpointLogsSeries {
					fmt.Fprint(w, `{"status":"success","data":[{"service":""}],"l9_result":{"partial":false}}`)
					return
				}
				fmt.Fprintf(w, `{"status":"success","data":{"result":[{"metric":{"value":"checkout","count":%d}}]},"l9_result":{"partial":false}}`, count)
			}))
			defer server.Close()
			contracts := testContracts(t, `[{"datasource":"prod","source":"logs","schema_version":1,"backend_limits":{"no_hidden_sampling":true,"max_rows":101}}]`)
			result, _, err := NewHandler(server.Client(), testConfig(server.URL), contracts)(context.Background(), nil, CatalogArgs{Datasource: "prod", Sources: []string{"logs"}, Protocol: "http", StartTimeISO: "2025-10-09T08:53:20Z", EndTimeISO: "2025-10-09T09:03:20Z", Include: []string{"services"}})
			if err != nil {
				t.Fatal(err)
			}
			var response CatalogResponse
			if err := json.Unmarshal([]byte(utils.GetTextContent(t, result)), &response); err != nil {
				t.Fatal(err)
			}
			if response.Result.Partial || len(response.Services) != 1 || response.Services[0].Count != count {
				t.Fatalf("want complete count %d; got services=%+v result=%+v", count, response.Services, response.Result)
			}
		})
	}
}

func TestReviewCatalogMissingBodyAttestationIsPartial(t *testing.T) {
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.URL.Path == constants.EndpointLogsSeries {
			fmt.Fprint(w, `{"status":"success","data":[{"service":""}],"l9_result":{"partial":false}}`)
			return
		}
		// Existing fixture intentionally has no l9_result envelope.
		json.NewEncoder(w).Encode(bodySampleResponse(`{"level":"info"}`))
	}))
	defer server.Close()
	contracts := testContracts(t, `[{"datasource":"prod","source":"logs","schema_version":1,"fields":{"level":"body"},"backend_limits":{"no_hidden_sampling":true,"max_rows":101}}]`)
	result, _, err := NewHandler(server.Client(), testConfig(server.URL), contracts)(context.Background(), nil, CatalogArgs{Datasource: "prod", Sources: []string{"logs"}, Protocol: "http", StartTimeISO: "2025-10-09T08:53:20Z", EndTimeISO: "2025-10-09T09:03:20Z", Include: []string{"fields"}})
	if err != nil {
		t.Fatal(err)
	}
	var response CatalogResponse
	if err := json.Unmarshal([]byte(utils.GetTextContent(t, result)), &response); err != nil {
		t.Fatal(err)
	}
	if !response.Result.Partial {
		t.Fatalf("unattested body sample reported complete: %+v", response.Result)
	}
}
```

nishantmodak (changes requested):
Reassessed the fresh review request on unchanged head `69a6c60f9a7c2fdcb1e88c762511b75ca1756d2e`. Both existing P2 blockers remain: [inventory count precision](https://github.com/last9/last9-mcp-server/pull/287#discussion_r4008243149) and [missing body-sample completeness attestation](https://github.com/last9/last9-mcp-server/pull/287#discussion_r4008243159). Preserve JSON integer counts before conversion, and mark the catalog partial when body-sample attestation is absent.

I reused the complete 30-file source inspection documented in [the September 14 review](https://github.com/last9/last9-mcp-server/pull/287#pullrequestreview-5201337537), which is on this exact head; this run rechecked the affected implementations/callers, current description, all discussions and CI. Retrieved that review's unchanged test attachment and ran it in a fresh isolated checkout using Go 1.27.1:

```sh
go test -p 1 ./internal/telemetry/catalog -run '^TestReviewCatalog' -count=2 -v
```

Both assertions fail on both runs: `9007199254740993` becomes `9007199254740992` with `partial=false`, and a body sample without `l9_result` is also reported complete. The existing catalog suite passes with those added regressions excluded. Current GitHub test and CodeQL checks are successful. The earlier broader race-suite results are historical evidence, not tests rerun here; this run used synthetic HTTP fixtures and did not exercise a live backend or deployment. The existing attachment remains the reproducer; no duplicate inline comments or test copies added.

PR #289, opened 2026-09-16
feat(grafana): five read-only Grafana dashboard tools under a grafana toolset

prathamesh-sonpatki:
## What
Add five read-only Grafana tools for the tenant MCP assistant, enabled under a new `grafana` toolset:

- `grafana_search_dashboards` — title substring search (`GET /api/search?query=&type=dash-db`)
- `grafana_get_dashboard` — filtered summary by default (version, tags, templating vars, each panel's type/datasource/gridPos/PromQL targets); `full_json=true` returns the raw Grafana payload
- `grafana_list_folders` — `GET /api/folders`
- `grafana_list_folder_dashboards` — resolves the folder's numeric id via `GET /api/folders/{uid}` then lists via `GET /api/search?folderIds=&type=dash-db`
- `grafana_list_datasources` — safe projection only; credential fields (`user`, `basicAuthUser`, `secureJsonFields`, …) are never forwarded to the model

`grafana_get_dashboard` surfaces unknown (plugin) panel types via `unsupportedPanelTypes` instead of dropping them.

## Why folder dashboards are two-step
`GET /api/folders/{uid}/dashboards` 404s through the Last9 Grafana proxy router (that Grafana API route is not registered), so the handler resolves the numeric folder id first and lists via the search endpoint. `search?folderUids=` is likewise unreliable through the proxy (the param is ignored). Verified live against `last9` org on app.last9.io.

## URLs
Tools hit the Grafana API proxy at `https://{api-host}/api/gp/v1/organizations/{org-slug}/api`, authenticated with the same tenant MCP token (`x-last9-api-token`). The `/api` edge prefix is required — without it the edge serves the SPA HTML.

## Live verification
All five tools exercised over stdio against the `last9` tenant: `list_folders` (8 folders), `search_dashboards` (query=service), `list_datasources` (projected rows), `list_folder_dashboards` on `ffgc0ffhiomwwa` (7 dashboards in GPU LLM Observability Stack), and `get_dashboard` on `fe4b48nrsnhfka` in both summary and full-json modes. All returned `isError=false` with real dashboard data.

## Tests
Unit tests use `httptest.NewServer`; `go test -race ./...` green.


internal/grafana/list_folder_dashboards.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		if folder.ID == 0 {
+			return nil, nil, fmt.Errorf("folder %q not found", args.FolderUID)
+		}
+
+		u := cfg.GrafanaAPIBaseURL + "/api/search?" + url.Values{
+			"folderIds": {strconv.Itoa(folder.ID)},
+			"type":      {"dash-db"},
+		}.Encode()
+		body, err := doJSONRequest(ctx, client, cfg, u)
```
nishantmodak: [P2] Expose or consume pagination before presenting the dashboard list

With 1,001 dashboards in a folder, this request retrieves only the first page and returns its 1,000 rows as an ordinary successful result. There is no continuation parameter or truncation marker, so the assistant cannot retrieve the missing dashboard through this tool and can incorrectly report a complete inventory. `grafana_search_dashboards` has the same problem for broad matches. Grafana's [documented search API](https://grafana.com/docs/grafana/latest/developer-resources/api-reference/http-api/api-legacy/folder_dashboard_search/) defaults to 1,000 results and uses `page` for subsequent results.

Required: either iterate pages with an explicit overall bound and report any truncation, or expose page/limit and continuation information through both tools. Raising the limit alone still silently loses results above that limit. Keep the folder/query filters on each page. The regression below exercises the actual handlers against that documented HTTP contract: both 1,001-row cases return only 1,000, twice; the 1-row and 1,000-row controls pass.

<details><summary>Executed native regression</summary>

Head `2a203a1a678cd81c37e25461477be2baecbdb648`; comparison `a1e5a17f400d90b46bf268fe04a48423a2d83a68` has no Grafana handlers, so this is new functionality checked against the API contract and positive controls, not a base/head regression claim.

Save as `internal/grafana/review_pagination_test.go`, then run from the repository root:
```sh
GOMAXPROCS=2 go test -p 1 -race ./internal/grafana -run TestReviewDashboardPagination -count=1 -timeout 45s -v
```
I ran this with Go 1.27.1 using this head's unchanged go.mod/go.sum and synthetic httptest servers only. Expected: all matching dashboards remain retrievable, without a silent incomplete success. Actual assertion: `silently omitted dashboards: got 1000, want 1001` for both search and folder handlers. The test represents an internal-pagination fix; an explicit continuation API should instead test retrieval across its exposed pages. Deployed proxy pagination was not contacted or independently tested.

```go
package grafana

import (
 "context"
 "encoding/json"
 "fmt"
 "net/http"
 "net/http/httptest"
 "strconv"
 "testing"
 "github.com/modelcontextprotocol/go-sdk/mcp"
)

func TestReviewDashboardPagination(t *testing.T) {
 for _, mode := range []string{"search", "folder"} {
  for _, count := range []int{1, 1000, 1001} {
   t.Run(fmt.Sprintf("%s/%d", mode, count), func(t *testing.T) {
    all := make([]SearchHit, count)
    for i := range all { all[i] = SearchHit{ID:i+1, UID:fmt.Sprintf("synthetic-%d", i), Title:"Synthetic service", Type:"dash-db"} }
    srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
     w.Header().Set("Content-Type", "application/json")
     if r.URL.Path == "/api/folders/f" { fmt.Fprint(w, `{"id":7,"uid":"f","title":"Synthetic"}`); return }
     if r.URL.Path != "/api/search" { http.NotFound(w,r); return }
     if mode == "folder" && r.URL.Query().Get("folderIds") != "7" { t.Error("lost folder filter") }
     limit, _ := strconv.Atoi(r.URL.Query().Get("limit")); if limit <= 0 || limit > 5000 { limit = 1000 }
     page, _ := strconv.Atoi(r.URL.Query().Get("page")); if page < 1 { page=1 }
     start := min((page-1)*limit, len(all)); end := min(start+limit, len(all))
     json.NewEncoder(w).Encode(all[start:end])
    }))
    defer srv.Close()
    cfg := testGrafanaConfig(srv.URL)
    var result *mcp.CallToolResult
    var err error
    if mode == "folder" { result, _, err = NewListFolderDashboardsHandler(srv.Client(),cfg)(context.Background(),nil,ListFolderDashboardsArgs{FolderUID:"f"})
    } else { result, _, err = NewSearchDashboardsHandler(srv.Client(),cfg)(context.Background(),nil,SearchDashboardsArgs{Query:"Synthetic"}) }
    if err != nil { t.Fatal(err) }
    var got []SearchHit
    if err := json.Unmarshal([]byte(result.Content[0].(*mcp.TextContent).Text),&got); err != nil { t.Fatal(err) }
    if len(got) != count { t.Fatalf("silently omitted dashboards: got %d, want %d",len(got),count) }
   })
  }
 }
}
```
</details>

nishantmodak (changes requested):
Reviewed `2a203a1a678cd81c37e25461477be2baecbdb648`: dashboard search and folder listing silently stop at the default first page. Please make the remaining results retrievable or explicitly report truncation; the inline finding includes the executed regression.

Inspected the complete 26-file diff, configuration/auth setup, registration/toolset filtering, HTTP handlers, projections, descriptions and tests. Existing Grafana, toolset, utils, auth and prompt suites passed locally with `-race -p 1`; focused root registration/dump tests also passed. Added probes passed for 30 seeded query-encoding cases (seed 289), two synthetic tenant-token configurations, credential-field projection, HTTP failures, response size limits and cancellation controlled by channels. Pagination failed at 1,001 rows for both handlers on two runs; 1 and 1,000-row controls passed. Current GitHub tests and CodeQL checks passed separately.

No live tenant, deployed proxy, external Grafana service or full repository suite was exercised. The local paging fixture follows Grafana's documented API contract; deployment-specific integration remains unverified.

prathamesh-sonpatki:
Addressed in `c512daa`.

`grafana_search_dashboards` and `grafana_list_folder_dashboards` now page through `/api/search` via `collectSearchHits` (`internal/grafana/http.go`): each page carries `limit=1000` plus the caller's query/folderId filters, walking pages until one comes back short or 5,000 rows are collected. Truncation is no longer silent — the response wrapper went from a bare array to `{"dashboards":[…], "truncated":bool}`, and `truncated: true` flags any result that hit the bound so the assistant reports an incomplete inventory instead of a complete one.

Pushed regression test `internal/grafana/pagination_test.go` mirroring your probe — 1, 1000 and 1001 dashboards each return fully for both handlers (1001 previously returned 1000), the folder filter is asserted on every page, and a 5,001-row run returns exactly 5,000 with `truncated: true`:

```
--- PASS: TestReviewDashboardPagination (search|folder x 1/1000/1001)
--- PASS: TestReviewDashboardPaginationTruncated
```

`go test -race -p 1 ./...` is clean. Tool descriptions and the CHANGELOG entry now document the wrapper and truncation flag.

nishantmodak (approved):
Reviewed all 27 changed files at `c512daa1eefaeeefdf6b7464f52ff387dd438b89`, including all five handlers, tenant configuration/authentication, HTTP failure handling, response projections, tool registration and callers. No remaining actionable blockers found.

The earlier pagination finding is fixed. I reran the attached regression with only its decoder adapted to the documented `dashboards` wrapper: both search and folder listings now return all 1,001 rows. Additional probes passed for 0/999/1,000/1,001/4,999/5,000/5,001/6,000 rows, exact ordering, truncation, preserved filters, later-page HTTP/malformed-response/cancellation failures, seeded query encoding, tenant headers and credential-field projection.

Validation: Go 1.27.1; focused Grafana/toolsets/utils/auth/prompts suites and probes with `-race -p 1`, plus root dump/registration tests. Current-head GitHub test and CodeQL checks are successful. Tests use synthetic HTTP servers; I did not exercise a deployed Grafana proxy or live tenant, or rerun the full integration suite.

